PR #10, opened 2022-03-02
IGVF-3-simplify-user-json

khzlin:


src/igvfd/schemas/user.json
```diff
@@ -35,7 +35,7 @@
             "title": "Lab",
             "description": "Lab user is primarily associated with.",
             "comment": "See lab.json for available identifiers.",
-            "permission": "import-items",
```
cricketsloan, line 38: @gabdank I don't remember catching this.  Was the permissions working on ENCODEd at all with this - versus _

src/igvfd/schemas/user.json
```diff
@@ -86,12 +86,30 @@
...
+                "enum": [
+                    "Primary Investigator",
+                    "Project Manager",
+                    "Submitter",
+                    "Post Doc",
+                    "Data Wrangler",
+                    "Scientist",
+                    "Computational Scientist",
+                    "Software Developer",
+                    "Other"
```
cricketsloan, line 102: @khzlin do we have a lot of graduate students or students as users?  I am also wondering what we call the NHGRI folks.  What do we call them now?

khzlin: I think "trainees" are used quite a lot to describe students a lot in calls. As for NHGRI folks, I'm not sure how to call them. Maybe "NHGRI Scientific Program Analyst"?

src/igvfd/schemas/user.json
gabdank: @cricketsloan we discovered that ENCODE allows "permission" to be ANY string. I just suggest to be consistent with other places using import items (import_items)

src/igvfd/schemas/user.json
gabdank: I would like to NOT use Trainees - due to a discussion of how to call the "young generation of researchers" that is happening at slack for IGVF

gabdank (commented):
@cricketsloan In general, I don't think we need upgrade scripts for these.. we are just creating the first version of Lab/User/Award

src/igvfd/schemas/user.json
cricketsloan: What about NHGRI Staff Member and is Student offensive?  I think we can add Post Doc.

cricketsloan:
Looks like adding a few more terms for people, updating the changelog to state where this broke from ENCODEd(This is version 1 of IGVF and it varies from version X of encodeD in the following ways.)

src/igvfd/schemas/user.json
khzlin: I like "NHGRI Staff Member". Is it OK to proceed without a term for Student for now? @gabdank @cricketsloan

cricketsloan: yes we can leave out student and get this done

cricketsloan (approved):

cricketsloan (approved):
Sorry, I had tried to review this from my phone previously, but apparently that did not save

keenangraham:
@khzlin Can you rebase/resolve conflicts when you get chance?

khzlin:
@keenangraham OK, rebased and tests passed.

PR #11, opened 2022-03-03
IGVF-2-simplify-lab

ian-whaling:


src/igvfd/tests/fixtures/schemas/lab.py
```diff
@@ -2,18 +2,21 @@
...
+        'pi': pi['@id'],
     }
     return testapp.post_json('/lab', item, status=201).json['@graph'][0]
 
 
 @pytest.fixture
-def lab_0_0():
+def lab_0_0(pi):
     return{
         'name': 'Fake Lab',
```
cricketsloan, line 17: I think that the convention that is being suggested above would require a hyphen and lower case for the name. "fake-lab" Unless this was a negative test.

src/igvfd/tests/test_views.py
```diff
@@ -163,23 +164,24 @@ def test_item_actions_filtered_by_permission(testapp, authenticated_testapp, lab
...
         'name': 'lab-a',
-        'title': 'The Lab A',
-        'address1': 'xyz'
+        'institute_label': 'Institute A',
+        'pi': pi['@id'],
     }
     item_url = testapp.post_json('/lab', initial).location
-
-    with execute_counter.expect(1):
+    print(item_url)
```
cricketsloan, line 174: There seems to be a print statement here that maybe was a debug?

src/igvfd/types/lab.py
```diff
@@ -22,3 +23,14 @@ class Lab(Item):
...
+    @calculated_property(
+            schema={
+                'title': 'Title',
+                'type': 'string',
+            }
+        )
+    def title(self, request, pi, institute_label):
+        pi_object = request.embed(pi, '@@object')
+        pi_name = pi_object.get('title')
+        return(pi_name + ', ' + institute_label)
```
cricketsloan, line 36: There is no end of line

cricketsloan (commented):
I have a few inline comments.  @gabdank I don't see a test for the calculated property.  In general, i am not worried, but I didn't know if you wanted it as an exercise.  Also, I have the same question as in the other review if this is going to be considered a pre-version 1 change as we have no data.  If so, we likely should document how this differs from ENCODEd somewhere, but maybe I am overthinking.

src/igvfd/tests/test_views.py
ian-whaling: Yes it was. Removed.

src/igvfd/types/lab.py
ian-whaling: Aded

keenangraham:
> I don't see a test for the calculated property.

Agree should add test.

src/igvfd/types/lab.py
keenangraham: Think should prefer f-strings where possible: 

`return f'{pi_name}, {institute_label}'`

keenangraham (changes requested):

ian-whaling:
I added a test in test_schemas.py, let me know if that is the correct place for the test and if you think the test should be more thorough.

keenangraham:
Should probably make a new file called `test_types_lab.py`. We have some analogous `test_types_*.py` in `encoded` for example: https://github.com/ENCODE-DCC/encoded/tree/dev/src/encoded/tests.

src/igvfd/tests/test_types_lab.py
```diff
@@ -1,4 +1,11 @@
...
+                            status=201).json['@graph'][0]
     assert(lab['title']) == 'Principal Investigator, Stanford'
-    assert(other_lab['title']) == 'IGVF VerifiedMember, Other Institute'
-    
\ No newline at end of file
+
+def test_bad_lab(testapp, pi):
+    testapp.post_json('/lab', {'name': 'bad-lab',
+                        'institute_label': 'Stanford'}, 
+                        status = 422)
```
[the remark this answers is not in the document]
ian-whaling: I created a lab with missing requirements, I thought running with 'status = 422' verified that there was an error. I modeled this test after the "bad" tests in https://github.com/IGVF-DACC/igvfd/blob/dev/src/igvfd/tests/test_download.py which don't assert anything

cricketsloan (dismissed):

cricketsloan:
> 

@keenangraham I believe that Ian made your changes, but I am blocked from approving.  Could you please check that this is to your satisfaction?

src/igvfd/tests/test_types_lab.py
```diff
@@ -0,0 +1,12 @@
+def test_title(testapp, pi):
+    lab = testapp.post_json('/lab', {'name': 'test-lab',
+                            'institute_label': 'Stanford',
+                            'pi': pi['@id']}, 
+                            status=201).json['@graph'][0]
+    assert(lab['title']) == 'Principal Investigator, Stanford'
+
```
keenangraham, line 7: Two lines.

keenangraham (dismissed):

keenangraham (approved):

PR #14, opened 2022-03-22
IGVF-55-linting

ottojolanki:
applied isort, autopep8, single-quote and end of line fixer


keenangraham:
I think we want to drop isort and maybe try adding JSON and YML checkers and maybe flake8?

cricketsloan:
I agree that a json formatter would be excellent.

ottojolanki:
added pretty-format-json.

keenangraham (dismissed):
Will need to rebase before merge.

forresttanaka:
I don’t know if this branch fixes this, but one thing I noticed with insert data is that prettier fixes those JSON files too, and it changed arrays to always have line breaks in them:
```
"prop": [
    "item1"
]
```
and it also corrected some incorrect indentation that was in the file (e.g. 7 spaces of indent instead of 8). I ended up having to turn off prettier for JSON files, but that seemed like a shame that I had to let go of the legitimate fixes it put in.

ottojolanki:
@forresttanaka  it should indent with 4 spaces and do "pretty formatting" so hopefully it is the same thing.

ottojolanki:
Could someone double check that `src/igvfd/tests/data/inserts/lab.json` and `src/igvfd/schemas/lab.json` are the way they need to be just in case.

cricketsloan:
Is there another competitor to prettier?  Or a version.  I see what Forrest is talking about, the breaking out of all of the all of the brackets is no "prettier"

        { "$ref": "mixins.json#/shared_status" }

is being changed to:

        {
            "$ref": "mixins.json#/schema_version"
        },

Possibly we live with this as it does find lots of other things

keenangraham (approved):
Think these are good changes, glad circle will run independent check.

I think the JSON multiline formatting is more consistent overall.

Sort of dragging Python back into repo dependency but seems disconnected enough from app (and only requires one time install?) that probably more useful than not.

Maybe should pin pip install: `pip install pre-commit==2.17.0`?

keenangraham (approved):

PR #22, opened 2022-04-07
IGVF-70-ontology-term-schema

jenjou:


cricketsloan (approved):
1) What were your thoughts on dbxrefs?  In the subclass or skipping them all together?

2) Will we have another ticket that addresses the mechanism for creating ontology.json?

jenjou:
> 1. What were your thoughts on dbxrefs?  In the subclass or skipping them all together?
> 2. Will we have another ticket that addresses the mechanism for creating ontology.json?

I left it out of the abstract class since the dbxrefs we allowed previously seemed to be specific to cell lines, but maybe it would make sense to have a placeholder for it, or allow other dbxrefs in.
I set up IGVF-75 for the code to generate ontology.json, but I haven't started working on it yet.

PR #25, opened 2022-04-12
fix-lint

keenangraham:


keenangraham:
Well I guess these aren't going to pass without antlr version fix. Will put these changes in that ticket.

PR #27, opened 2022-04-15
IGVF-68 New properties of sample, cell_lines, and tissue schema.

KAndreeva:
Initial implementation of new properties of sample, cell_lines, and tissue schema.

- nih_institutional_certification (added at the level of biosample use encodeD as a model) abstract class.  This is inherited by tissue and cell_lines classes.
- internal_tags (added in the mixins and then on the level of sample, use encodeD as a model but limit the tags).  This is inherited by tissue, cell_lines, and technical_sample classes.
- Added unit test for the new properties.

### DIFF
--- a/src/igvfd/schemas/mixins.json
+++ b/src/igvfd/schemas/mixins.json
@@ -299,5 +299,26 @@
             "pattern": "^(\\S+(\\s|\\S)*\\S+|\\S)$",
             "formInput": "textarea"
         }
+    },
+    "internal_tags": {
+        "internal_tags": {
+            "title": "Internal tags",
+            "description": "Some samples are part of particular data collections.",
+            "comment": "Do not submit. Internal tags are for DCC use only.",
+            "type": "array",
+            "default": [],
+            "permission": "import_items",
+            "uniqueItems": true,
+            "items": {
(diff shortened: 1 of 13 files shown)


cricketsloan (commented):
In Encode, I remembered internal tags being groupings like GTex and ENCORE.  The list of types of investigation is interesting.  How will it work?  Will we be tagging a biosample with what it was used to investigate?

src/igvfd/schemas/mixins.json
```diff
@@ -299,5 +299,26 @@
...
+            "description": "Some samples are part of particular data collections.",
+            "comment": "Do not submit. Internal tags are for DCC use only.",
+            "type": "array",
+            "default": [],
+            "permission": "import_items",
+            "uniqueItems": true,
+            "items": {
+                "type": "string",
+                "enum": [
+                    "Enhancers",
```
cricketsloan, line 315: We have decided that we will restrict to ENCODE only at this point

cricketsloan (changes requested):

ian-whaling (approved):

ian-whaling (approved):

PR #29, opened 2022-04-18
IGVF-83-load-gz-ontology-json

keenangraham:


src/igvfd/ontology.py
```diff
@@ -0,0 +1,33 @@
...
+
+from pathlib import Path
+
+
+def includeme(config):
+    config.scan(__name__)
+    config.registry['ontology'] = load_ontology()
+
+
+ONTOLOGY_NAME = 'ontology.json.gz'
```
ottojolanki, line 12: nit, but I suppose `ONTLOGY_FILE_NAME` would be more appropriate.

keenangraham: Updated.

ottojolanki (approved):
LGTM

PR #30, opened 2022-04-21
IGVF-78-sample-ontology-term

jenjou:


src/igvfd/schemas/sample_ontology_term.json
```diff
@@ -0,0 +1,64 @@
...
+        }
+    ],
+    "properties": {
+        "schema_version": {
+            "default": "1"
+        },
+        "term_id": {
+            "@type": "@id",
+            "title": "Ontology ID",
+            "description": "An ontology identifier describing a biological sample, assay, trait, or disease.",
```
cricketsloan, line 43: This description seems like it would be on the level of ontology not at the level of sample_ontology

cricketsloan (commented):
I was expecting to see an abstract ontology_term schema.

src/igvfd/schemas/sample_ontology_term.json
jenjou: Will fix this.

src/igvfd/tests/test_schema_sample_ontology_term.py
```diff
@@ -0,0 +1,9 @@
+import pytest
+
+
+def test_sample_ontology_term_term_id_regex(sample_ontology_term_1, testapp):
+    res = testapp.patch_json(
+        sample_ontology_term_1['@id'],
+        {'term_id': 'ABC:12345'},
+        expect_errors=True)
+    assert(res.status_code == 422)
```
cricketsloan, line 9: Should we have a test that checks that when you put in some known cell that you are getting reasonable slims upon posting?

src/igvfd/schemas/sample_ontology_term.json
```diff
@@ -0,0 +1,64 @@
...
+            "$ref": "mixins.json#/aliases"
+        },
+        {
+            "$ref": "mixins.json#/standard_status"
+        },
+        {
+            "$ref": "mixins.json#/notes"
+        },
+        {
+            "$ref": "ontology_term.json#/properties"
```
cricketsloan, line 33: Sorry for my earlier comment.  I was not seeing this line as I was looking at the top of the list.

cricketsloan (commented):
I think we need more testing around the slimming to at least make sure that something is getting calculated.  

I also wonder if you want this to be biosample and not sample as we don't know what ontology we might use for sample.  However, i am fine leaving it as sample.
  
Other than that, I think this works.

cricketsloan (approved):
Thank you for adding the tests

PR #36, opened 2022-04-25
IGVF-47-treatment-schema-2

meenakshikagda:


src/igvfd/schemas/treatment.json
```diff
@@ -0,0 +1,206 @@
+{
+    "title": "Treatment",
+    "id": "/profiles/treatment.json",
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "type": "object",
+    "required": [
```
gabdank, line 6: I am not sure this is a reasonable list of required properties. I am not sure duration should be required. If the amount is required then the units should be required as well, otherwise, you are leaving it to the user to double guess why their submission is not valid, and require them to read dependencies?

src/igvfd/types/treatment.py
```diff
@@ -0,0 +1,31 @@
...
+    item_type = 'treatment'
+    schema = load_schema('igvfd:schemas/treatment.json')
+
+    @calculated_property(
+        schema={
+            'title': 'Title',
+            'type': 'string',
+        }
+    )
+    def title(self, treatment_term_name, amount, amount_units, duration, duration_units):
```
gabdank, line 28: this calculated property will need to change if we drop duration from the list of required properties.

src/igvfd/types/treatment.py
```diff
@@ -0,0 +1,31 @@
...
+        text = 'Treated with {} {} {} for {} {}.'.format(
```
gabdank, line 29: I think you should also add a test that makes sure the calculated property is calculated as expected

src/igvfd/types/treatment.py
```diff
@@ -0,0 +1,31 @@
...
+            amount, amount_units, treatment_term_name, duration, duration_units)
+        return(text)
```
gabdank, line 31: I think it should be "return text" and not "return(text)"

src/igvfd/schemas/treatment.json
```diff
@@ -0,0 +1,206 @@
...
+        }
+    },
+    "properties": {
+        "schema_version": {
+            "default": "1"
+        },
+        "documents": {
+            "description": "Documents that describe the treatment protocol."
+        },
+        "amount": {
```
gabdank, line 93: please add description

src/igvfd/schemas/treatment.json
```diff
@@ -0,0 +1,206 @@
...
+            "title": "Amount",
+            "type": "number"
+        },
+        "amount_units": {
```
gabdank, line 97: please add description, and probably comment explaining that units go with amount

src/igvfd/schemas/treatment.json
gabdank: and comment explaining that specification of units is required

src/igvfd/schemas/treatment.json
```diff
@@ -0,0 +1,206 @@
...
+                "nM",
+                "percent",
+                "μg/kg",
+                "μg/kg",
+                "μg/mL",
+                "μM"
+            ]
+        },
+        "duration": {
+            "title": "Time duration",
```
gabdank, line 114: please add description and comment explaining that units must be specified

src/igvfd/schemas/treatment.json
```diff
@@ -0,0 +1,206 @@
...
+            "type": "number"
+        },
+        "duration_units": {
+            "title": "Time duration units",
```
gabdank, line 118: please add description and comment explaining that duration needs to be specified

src/igvfd/schemas/treatment.json
```diff
@@ -0,0 +1,206 @@
...
+                "agonist",
+                "antagonist",
+                "control",
+                "differentiation",
+                "de-differentiation",
+                "perturbation",
+                "selection",
+                "stimulation"
+            ],
+            "comment": "The intended purpose for treating the samples; Activation: treatment is known to activate a pathway in the biosample; Agonist: a substance which initiates a physiological response when combined with a receptor; Antagonist: a substance that interferes with or inhibits the physiological action of another; Control: treatment applied to a sample for control purposes; Differentiation: treatment that is applied to convert a less specialized cell to a mature cell; De-differentiation: treatment used to reprogram differentiated cells back to less determined cell states; Perturbation: treatment applied to the sample in order to study the effect of its application; Selection: treatment used to affect biosample in a way that can be used to distinguish cells and selected for in the downstream steps; Stimulation: treatment stimulates a cellular pathway."
```
gabdank, line 145: (1) Differentiation: treatment that is applied to convert a less specialized cell to a mature cell - I think should change to: `Differentiation: treatment that is applied to convert a less specialized cell to a more specialized cell`.
(2) Selection: treatment used to affect biosample in a way that can be used to distinguish cells and selected for in the downstream steps; should change into `Selection: treatment used to affect biosample in a way that can be used to distinguish cells and select for in the downstream steps;`. 
(3) Stimulation: treatment stimulates a cellular pathway. should be: `Stimulation: treatment apllied to stimulate a cellular pathway.`

src/igvfd/schemas/treatment.json
```diff
@@ -0,0 +1,206 @@
...
+        },
+        "temperature (Celsius)": {
+            "title": "Temperature",
+            "type": "number",
+            "comment": "Temperature can only be specified in Celsius"
+        },
+        "post_treatment_time": {
```
gabdank, line 152: add a dependency for post treatment time and units (they should go together), add a test for this dependency

src/igvfd/schemas/treatment.json
```diff
@@ -0,0 +1,206 @@
...
+                "minute",
+                "hour",
+                "day",
+                "week",
+                "month"
+            ]
+        },
+        "treatment_type": {
+            "title": "Treatment type",
+            "description": "The classification of the treatment.",
```
gabdank, line 174: add a comment giving a short example of chemical and protein treatments.

src/igvfd/schemas/treatment.json
```diff
@@ -0,0 +1,206 @@
...
+        "treatment_term_id": {
+            "@type": "@id",
+            "title": "Biosample treatment term ID",
+            "description": "Ontology identifier describing a component in the treatment.",
+            "type": "string",
+            "pattern": "^((CHEBI:[0-9]{1,7})|(UniProtKB:[A-Z0-9]{6})|(NTR:[0-9]{2,8}))$"
+        },
+        "treatment_term_name": {
+            "title": "Biosample treatment",
+            "description": "Ontology term describing a component in the treatment.",
```
gabdank, line 190: add a comment providing example of ontological terms for chemical and protein treatments

gabdank (changes requested):
please see my comments

src/igvfd/schemas/treatment.json
```diff
@@ -0,0 +1,206 @@
...
+        {
+            "$ref": "mixins.json#/product_info"
+        }
+    ],
+    "dependentSchemas": {
+        "amount": {
+            "required": [
+                "amount_units"
+            ],
+            "comment": "amount_units must be specified if amount is specified"
```
gabdank, line 49: both properties are required, I think this part could be dropped

src/igvfd/schemas/treatment.json
```diff
@@ -0,0 +1,206 @@
...
+        },
+        "duration": {
+            "required": [
+                "duration_units"
+            ],
+            "comment": "duration_units must be specified if duration is specified"
```
gabdank, line 55: I think this dependency means that one can specify duration_units, but not duration? I think you need to add the other "direction". If the duration is specified you have to have units, and if you have units you have to have duration.

gabdank: same for post_treatment_units and post_treatment_time

cricketsloan: @meenakshikagda you can use the dependancies in Ian's PR for Organiod as an example

src/igvfd/tests/test_schema_treatment.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def test_treatment_amount_dependency(testapp):
+    item = {
+        'treatment_term_name': 'lactate',
+        'treatment_term_id': 'CHEBI:24996',
+        'treatment_type': 'chemical',
+        'amount': '10'
+    }
+    return testapp.post_json('/treatment', item, expect_errors=True)
```
cricketsloan, line 11: This return statement means that lines 12-19 never get executed.  You will need to change the line to res = like in line 23.  However, this test is of a dependency that is redundant.  You do not need to ensure that amount units is required if amount is there, it is always required.  So, if amount and amount_units are required, this test from lines 4-19 is not required.

src/igvfd/tests/data/inserts/treatment.json
```diff
@@ -0,0 +1,28 @@
+[
+    {
```
cricketsloan, line 2: I would like to see another 2-3 treatments, one that has an NTR, ones using different purposes, one using notes.  The inserts provide you with a way to test that all of your properties are working and gives you reasonable examples to like to biosamples.

src/igvfd/schemas/treatment.json
```diff
@@ -0,0 +1,206 @@
...
+            "$ref": "mixins.json#/product_info"
+        }
+    ],
+    "dependentSchemas": {
+        "amount": {
+            "required": [
+                "amount_units"
+            ],
+            "comment": "amount_units must be specified if amount is specified"
+        },
```
cricketsloan, line 50: This dependency can be dropped as it is redundant to the required field above.  It is like saying I require childcare every day of the week AND if it is Thursday, I also require childcare.  If you drop lines 45-50 then you will also be able to drop the test below.

src/igvfd/schemas/treatment.json
```diff
@@ -0,0 +1,206 @@
...
+                "day",
+                "week",
+                "month"
+            ],
+            "comment": "post_treatment_time_units should be used in conjunction with post_treatment_time."
+        },
+        "temperature (Celsius)": {
+            "title": "Temperature",
+            "type": "number",
+            "description": "Temperature can only be specified in Celsius"
```
cricketsloan, line 180: Line 180 is an example of mixed description and comment.  Below I have split the concepts.
"description":"The temperature in Celsius to which the sample was exposed"
"comment":"Temperature must be converted to Celsius for submission."

cricketsloan (changes requested):

src/igvfd/schemas/treatment.json
meenakshikagda: I love the analogy. Made me smile 😃

gabdank (changes requested):
@meenakshikagda I think the PR has to be fixed, as currently, it includes schema changes from your branch and from other branches mixed together.

src/igvfd/loadxl.py
```diff
@@ -18,7 +18,7 @@
     'source',
     'lab',
     'document',
-    'sample_ontology_term',
```
cricketsloan, line 21: I don't think you meant to get rid of this line.

cricketsloan (changes requested):
I think this is a reasonable draft to start to integrate with the other work.  There is one line (marked inline) that I do not think you meant to erase.  Once you resolve that line this will be ready to merge.

cricketsloan (approved):

PR #37, opened 2022-04-26
IGVF-80-use-$merge-in-dependencies

ian-whaling:


cricketsloan (changes requested):
1) I don't think that you need to update the version as there is no backwards compatibility issue happening here.  The end schema is the same.

2) Would you be able to have the sample/dependant-schema inherited as well as the biosample/dependent-schema.  For instance, the part about lot requiring product_id should be at the level of sample and inherited in technical sample as well as biological sample

gabdank (changes requested):
(1) I think that next time you should not name the branch "IGVF-80-"  if possible provide at least some context in the name. Otherwise at the point we spin up demos - nobody will understand what your branch is supposed to be doing without reading the ticket.
(2) I don't understand why you haven't used $merge in technical sample?

ian-whaling:
> (1) I think that next time you should not name the branch "IGVF-80-" if possible provide at least some context in the name. Otherwise at the point we spin up demos - nobody will understand what your branch is supposed to be doing without reading the ticket. (2) I don't understand why you haven't used $merge in technical sample?

Yes sorry about the name, I named the branch IGVF-80-$merge-implementation and evidently using $ causes problems in naming. I added to technical sample.

src/igvfd/schemas/technical_sample.json
```diff
@@ -65,12 +65,9 @@
...
     ],
     "dependentSchemas": {
-        "lot_id": {
-            "required": [
-                "product_id"
-            ],
-            "comment": "Technical sample with lot_id must include product_id."
-        }
+        "$merge": [
+            "biosample.json#/dependentSchemas"
```
cricketsloan, line 69: I think that this would be sample.json#/dependentSchemas

cricketsloan (changes requested):
Technical should $merge from sample not biosample

cricketsloan (approved):

gabdank (approved):

PR #41, opened 2022-04-28
IGVF-84-assay-ontology-term

jenjou:


src/igvfd/tests/data/inserts/assay_ontology_term.json
```diff
@@ -0,0 +1,20 @@
+[
+    {
+        "term_id": "OBI:0002041",
+        "term_name": "STARR-seq",
+        "status": "released",
+        "uuid": "df207aeb-3484-3730-b64c-d3fdb4506a5e"
```
cricketsloan, line 6: Do you want to have an insert that uses aliases and notes since those are fields that are included?  I think that makes sense for future developers to have an example of every field.

cricketsloan (commented):
@jenjou This looks good.  I have a few ideas, but if nothing very major.

src/igvfd/tests/data/inserts/assay_ontology_term.json
jenjou: I think that makes sense, will add this

src/igvfd/loadxl.py
```diff
@@ -22,6 +22,7 @@
     'cell_line',
     'tissue',
     'technical_sample',
+    'assay_ontology_term',
```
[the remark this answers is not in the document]
jenjou: I was unsure how we want to sort this. I placed it at the end because technically, none of the preceding objects utilize assay_ontology_term, but I realize it looks somewhat out of place being separated from sample ontology. Do we have a preference here?

cricketsloan (approved):
Looks good

PR #42, opened 2022-04-28
adding myself to the user.json test insert file

dougli1sqrd:
Is this fine for the UUID? I just generated one and placed it in


hitz (changes requested):
I think you want to be in group "admin" as well.

gabdank (commented):
Eric, please in the future name your PRs starting with the ticket number (similarly to the branch name)

hitz (approved):

PR #48, opened 2022-05-06
adding primary cells

cricketsloan:


src/igvfd/schemas/changelogs/primary_cells.md
```diff
@@ -0,0 +1 @@
+## Changelog for primary_cell.json
```
ian-whaling, line 1: should be primary_cells.json

ian-whaling (changes requested):
Looks good. 

Maybe I would suggest not pluralizing this class (all instances of "primary_cells" -> "primary_cell")? I think classes in general should be singular as they represent a single object and I don't see another example of a pluralized class. But maybe you disagree.

In this case the header in the changelog would remain unchanged so you can dismiss that change request.

cricketsloan:
I debated a lot about calling it primary cell versus primary cells.  It makes more sense in english to say you have a sample of primary cells but in this context I think you are right.  I will fix

ian-whaling (approved):

PR #50, opened 2022-05-06
IGVF-111-disease-ontology

jenjou:


src/igvfd/tests/test_schema_disease_ontology_term.py
```diff
@@ -0,0 +1,9 @@
+import pytest
+
+
+def test_disease_ontology_term_term_id_regex(disease_ontology_term_alzheimers, testapp):
```
gabdank, line 4: I think you can also add a test for required properties
I also think the inverse of regex failure could be executed (even though inserts cover that)

gabdank (changes requested):

gabdank (approved):

PR #51, opened 2022-05-09
IGVF-87-add-gene

meenakshikagda:


src/igvfd/tests/data/inserts/gene.json
```diff
@@ -0,0 +1,116 @@
...
+    {
+        "geneid": "2995",
+        "uuid": "83a311d1-7ed3-41bb-85f0-1f00c458e6fc",
+        "dbxrefs": [
+            "MIM:110750",
+            "HGNC:4704",
+            "ENSEMBL:ENSG00000136732",
+            "Vega:OTTHUMG00000131464",
+            "UniProtKB:P04921",
+            "RefSeq:NG_007479"
```
cricketsloan, line 33: Could you add a genecards reference to test the regex

cricketsloan: Actually we should have one for each regex, so there are a few more I am realizing

src/igvfd/tests/data/inserts/gene.json
```diff
@@ -0,0 +1,116 @@
...
+            "ENSEMBL:ENSMUSG00000093769",
+            "Vega:OTTMUSG00000021930",
+            "UniProtKB:P84228",
+            "UniProtKB:B9EI85"
+        ],
+        "name": "histone cluster 2, H3c1",
+        "ncbi_entrez_status": "live",
+        "organism": "Mus musculus",
+        "status": "released",
+        "symbol": "Hist2h3c1",
```
cricketsloan, line 66: Could you add note as that is the only property in the schema that is not exercised in the inserts.

cricketsloan (commented):
This is good.  The fly and worm details have been stripped and we have switched to organism as an enum.  

The original inserts did not cover all of the properties and all of the regexs, could you please add those?

cricketsloan (approved):
For your extensive test of the location dependency, you might have used a for loop to clean up that code (for each assembly).  But since there are only 2 of each assembly, it doesn't make sense to redo this until there are 3 assemblies.

PR #54, opened 2022-05-10
IGVF-86 docs

gabdank:


src/igvfd/schemas/changelogs/document.md
```diff
@@ -0,0 +1,26 @@
...
+    "enum": [
+        "FACS",
+        "immunoblot",
+        "immunofluorescence",
+        "immunoprecipitation",
+        "mass spectrometry",
+        "PCR",
+        "restriction digest",
+        "RT-qPCR",
+        "Sequencing"
```
jenjou, line 25: To be consistent with the other enums, "sequencing" could be lower case.

jenjou:
There was no test_schema_documents created when we initially added the Documents schema, so perhaps we should add that in this ticket. The fixture experimental_protocol_document is also unused by any tests.

src/igvfd/schemas/changelogs/document.md
gabdank: ok

jenjou (approved):
Test was added and other comments addressed.

PR #58, opened 2022-05-14
IGVF-126-biosample-age-string

ian-whaling:
Encode has [dependencies](https://github.com/ENCODE-DCC/encoded/blob/428818fced4c5ca049680cf5143966ab46dd987d/src/encoded/schemas/human_donor.json#L46) for age and life_stage and a pattern for [age](https://github.com/ENCODE-DCC/encoded/blob/428818fced4c5ca049680cf5143966ab46dd987d/src/encoded/schemas/human_donor.json#L120) let me know if I should implement these. @cricketsloan

### DIFF
--- a/src/igvfd/schemas/biosample.json
+++ b/src/igvfd/schemas/biosample.json
@@ -167,8 +167,7 @@
         "age": {
             "title": "Age",
             "description": "Age of organism at collection of the biosample.",
-            "type": "number",
-            "minimum": 0
+            "type": "string"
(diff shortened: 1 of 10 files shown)


jenjou (changes requested):
Code looks good to me, but I think we can add an insert using the "90 or above" regex pattern.

jenjou (dismissed):

jenjou (approved):

PR #63, opened 2022-05-18
IGVF-137-aliases-fix

ian-whaling:
Let me know if you think we should add any other patterns (non-lab patterns such as encode) @gabdank 
https://github.com/ENCODE-DCC/encoded/blob/dev/src/encoded/schemas/mixins.json#L67


ian-whaling:
add igvf and igvf-dacc prefixes

gabdank (dismissed):

gabdank (dismissed):

gabdank (approved):

PR #67, opened 2022-05-24
IGVF-144-rework-related-biosamples

ian-whaling:


src/igvfd/loadxl.py
```diff
@@ -25,9 +25,9 @@
     'human_donor',
     'rodent_donor',
     'treatment',
+    'tissue',
```
cricketsloan, line 28: why did this move?

src/igvfd/schemas/biosample.json
```diff
@@ -224,31 +224,25 @@
...
-            "comment": "A list of biosamples should all have the same relationship to this biosample.",
+        "pooled_from": {
+            "title": "Biosample(s) pooled from",
+            "description": "The biosample(s) this biosample is pooled from.",
             "type": "array",
+            "default": [],
             "uniqueItems": true,
             "items": {
-                "description": "A related biosample",
-                "comment": "See biosample.json for available identifiers.",
```
cricketsloan, line 235: I would leave the "comment"

src/igvfd/schemas/biosample.json
```diff
@@ -224,31 +224,25 @@
...
+            "comment": "For other specific relationships such as differentiation or hosting there are separate properties.",
             "type": "string",
-            "enum": [
-                "originated_from",
-                "pooled_from",
-                "separated_from",
-                "hosted_by",
-                "derived_from"
-            ]
+            "linkTo": "Biosample"
```
cricketsloan, line 245: I wonder if we should make this generic to Sample instead of tied to biosample.

src/igvfd/tests/data/inserts/differentiated_tissue.json
```diff
@@ -13,6 +13,7 @@
         "post_differentiation_time_units": "day",
         "treatments": [
             "4fbb0dc2-c72e-11ec-9d64-0242ac120002"
-        ]
+        ],
+        "differentiated_from": "/cell-lines/467c72a2-4f84-2c8f-96b0-ec8715e18185/"
```
cricketsloan, line 17: OK, this is a big question.  Thinking about the samples you were trying to post, do you think that they will want to register the specific H9 sample or do you think that this should point to an ontology term like H9.

cricketsloan (changes requested):
There are a couple of small changes.  But there is a big discussion point of whether differentiated_from should link to biosample or to biosample_ontology.

src/igvfd/loadxl.py
ian-whaling: so primary_cell can point to tissue when loading the inserts

src/igvfd/tests/data/inserts/differentiated_tissue.json
ian-whaling: I think encode points to a specific H9 sample that all use the same donor. This cell line sample may have some info that can't be provided on the subsequent differentiated sample? Thinking about it I don't really see the point of this redundancy though, but someone with more experience on encode would probably provide more insight. I think it does make more sense to point to the ontology term for the derived cell line.

src/igvfd/schemas/biosample.json
ian-whaling: I don't see the harm in making it an option if you think it is a feasible a biosample may be a part of a technical sample.

src/igvfd/tests/data/inserts/differentiated_tissue.json
cricketsloan: Let's ask @jenjou and @gabdank I am thinking that we should consider differentiated from as an ontology term rather than a specific biosample, but I think those two would have more recent experience.  Also, I think if we add host in the future we should have that be an ontology term as well.

jenjou: Linking to a specific biosample could allow the labs to indicate additional metadata such as the starting quantity of cells, or provide a way to link a small group of biosamples originating from the same starting cells and differentiated into different cell fates. However, I'm not sure how often users would search for that type of info.

src/igvfd/loadxl.py
gabdank: that will break once we hook in the other direction (primary_cell -> tissue will work now, but (cell_line -> primary_cell will not). It should be handled in 
PHASE1_PIPELINES = {
    'user': [
        remove_keys('lab', 'submits_for'),
    ],
} instead

src/igvfd/tests/data/inserts/differentiated_tissue.json
gabdank: i think we should link to biosample, rather than ontology. linking to ontology suffers from the downside cellosaurus suffers from - you are basically required to register as ontology term every genetically modified cell line, which I think we want to avoid (unlike cellosaurus).  i don't think differentiated_from is a required property and should be used to point to the sample it was differentiated from if such is available

src/igvfd/schemas/biosample.json
gabdank: i am leaning towards leaving it Biosample for the reason Ian brings up.

src/igvfd/loadxl.py
cricketsloan: @ian-whaling I think this detail about moving this logic to PHASES is what is left.  Do you need more clarity on this request or have you figured it out?

src/igvfd/tests/data/inserts/primary_cell.json
```diff
@@ -62,6 +63,9 @@
         "source": "/labs/j-michael-cherry/",
         "donors": [
             "j-michael-cherry:alias_human_donor1"
+        ],
+        "pooled_from": [
+            "/primary-cells/578c72a2-4f84-2c8f-96b0-ec8715e18185"
```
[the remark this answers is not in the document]
cricketsloan: "minItems": 2

cricketsloan (dismissed):
I am surprised that a few of the lines in your new work pass our style checker. I also realized that we might want to set pooled_from to have a minimum of 2 items.  However, i am OK leaving that for another fix.

src/igvfd/loadxl.py
```diff
@@ -511,6 +525,21 @@ def get_pipeline(testapp, docsdir, test_only, item_type, phase=None, method=None
...
+    ],
+    'tissue': [
+        skip_rows_missing_all_keys('biosample', 'pooled_from'),
+        skip_rows_missing_all_keys('biosample', 'part_of'),
+    ],
+    'primary_cell': [
+        skip_rows_missing_all_keys('biosample', 'pooled_from'),
+        skip_rows_missing_all_keys('biosample', 'part_of'),
+    ],
+    'differentiated_cell': [skip_rows_missing_all_keys('biosample', 'differentiated_from'), skip_rows_missing_all_keys('biosample', 'pooled_from'), skip_rows_missing_all_keys('biosample', 'part_of'), ],
```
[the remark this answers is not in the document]
ottojolanki: we are not using black. just linters.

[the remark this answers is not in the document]
ottojolanki: There are multiple ways to indent such that `autopep8` accepts it.

cricketsloan (dismissed):
The linter question is bigger than this ticket.  So we will leave that out of the discussion.

cricketsloan (approved):
I am sorry that I did not catch the loadxl problem last time.  These changes look good.

cricketsloan (approved):

PR #75, opened 2022-06-08
IGVF-175-organism-mixin

ian-whaling:


gabdank (changes requested):
I think you are allowing to create human donor that is hooked into mus musculus organism? Probably should avoid that.
I think you should exercise the restriction of enum list of organisms. For example, try posting a human donor with non human organism

gabdank (approved):

PR #83, opened 2022-06-21
IGVF-51-publications

ian-whaling:


src/igvfd/schemas/changelogs/publication.md
```diff
@@ -0,0 +1,3 @@
+## Changelog for publication.json
+
+### Minor changes since version 1
```
gabdank, line 3: I don't think we should put the line about minor changes if there were no changes

src/igvfd/schemas/publication.json
```diff
@@ -0,0 +1,135 @@
+{
+    "title": "Publication",
+    "description": "Schema for a publication page.",
```
gabdank, line 3: I think it is not publication page, but publication object

src/igvfd/schemas/publication.json
```diff
@@ -0,0 +1,135 @@
...
+                "title": "Published by",
+                "type": "string",
+                "enum": [
+                    "community",
+                    "IGVF",
+                    "ENCODE"
+                ]
+            }
+        },
+        "publication_page": {
```
gabdank, line 127: I think we should drop it until we have pages editable by submitters

src/igvfd/schemas/publication.json
```diff
@@ -0,0 +1,135 @@
+{
+    "title": "Publication",
+    "description": "Schema for a publication page.",
+    "$id": "/profiles/publication.json",
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "type": "object",
+    "required": [
+        "title",
```
gabdank, line 8: i think "identifiers" should be required

src/igvfd/schemas/publication.json
```diff
@@ -0,0 +1,135 @@
...
+            "title": "Identifiers",
+            "description": "The identifiers that reference data found in the object.",
+            "type": "array",
+            "uniqueItems": true,
+            "items": {
+                "title": "Identifier",
+                "description": "An identifier that references data found in the object.",
+                "type": "string",
+                "uniqueKey": "publication:identifier",
+                "pattern": "^(PMID:[0-9]+|doi:10\\.[0-9]{4}[\\d\\s\\S\\:\\.\\/]+|PMCID:PMC[0-9]+|[0-9]{4}\\.[0-9]{4})$"
```
gabdank, line 107: I think the pattern should support whatever identifiers are in BioRxiv (and maybe it is already, just make sure)

src/igvfd/schemas/publication.json
```diff
@@ -0,0 +1,135 @@
...
+            "type": "string"
+        },
+        "journal": {
+            "title": "Journal",
+            "description": "The journal of the publication.",
+            "type": "string"
+        },
+        "identifiers": {
+            "title": "Identifiers",
+            "description": "The identifiers that reference data found in the object.",
```
gabdank, line 99: I think the list of identifiers should have minItem=1, then the requirement to have this list would make sense

src/igvfd/tests/fixtures/schemas/publication.py
```diff
@@ -0,0 +1,11 @@
+import pytest
+
+
+@pytest.fixture
+def publication(testapp, lab, award):
+    item = {
```
gabdank, line 6: if we make identifiers required will need to be updated

src/igvfd/tests/test_schema_publication.py
```diff
@@ -0,0 +1,12 @@
+import pytest
+
+
+def test_date_published(publication, testapp):
+    res = testapp.patch_json(
```
gabdank, line 5: add test that makes sure identifiers are unique

src/igvfd/tests/test_schema_publication.py
```diff
@@ -0,0 +1,12 @@
...
+        publication['@id'],
+        {'date_published': '2022-2-2'}, expect_errors=True)
```
gabdank, line 7: add test for required properties
and test for regex of identifiers

gabdank (changes requested):

src/igvfd/schemas/publication.json
ian-whaling: BioRxiv uses doi, so it is supported. I put an insert to reflect this.

src/igvfd/types/publication.py
```diff
@@ -0,0 +1,39 @@
...
+        keys = super(Publication, self).unique_keys(properties)
+        if properties.get('identifiers'):
+            keys.setdefault('alias', []).extend(properties['identifiers'])
+        return keys
+
+    @calculated_property(condition='date_published', schema={
+        'title': 'Publication year',
+        'type': 'integer',
+    }
+    )
```
gabdank, line 33: please put these two on the same line like line 18

src/igvfd/schemas/publication.json
```diff
@@ -0,0 +1,131 @@
...
+                    "format": "date"
+                },
+                {
+                    "comment": "Historical date formats, not valid for new submissions.",
+                    "permission": "import_items",
+                    "pattern": "^[0-9]{4}[^-]*$"
+                }
+            ]
+        },
+        "date_revised": {
```
gabdank, line 66: we discussed it at a wranglers meeting, do not use  date-time for any of the date properties

only timestamps are date-time

src/igvfd/schemas/publication.json
```diff
@@ -0,0 +1,131 @@
...
+            "type": "string",
+            "oneOf": [
+                {
+                    "format": "date"
+                },
+                {
+                    "comment": "Historical date formats, not valid for new submissions.",
+                    "permission": "import_items",
+                    "pattern": "^[0-9]{4}[^-]*$"
+                }
```
gabdank, line 63: what is the purpose of these?

src/igvfd/types/publication.py
```diff
@@ -0,0 +1,39 @@
...
+    def publication_year(self, date_published):
+        likely_year = date_published[:4]
```
gabdank, line 35: I think you should use python date conversions, something like: 
```
dt = datetime.strptime(datestring, "%Y-%m-%d")
print (dt.year)
```

gabdank (changes requested):

src/igvfd/schemas/publication.json
ian-whaling: Not sure, this was ported from ENCODE but it looks to have been deprecated there so I will remove it.

src/igvfd/tests/test_types_publication.py
```diff
@@ -0,0 +1,8 @@
+import pytest
+
+
+def test_publication_year(testapp, publication):
```
gabdank, line 4: can you please add a test to make sure the publication without date_published works ok?

gabdank (changes requested):

src/igvfd/types/publication.py
```diff
@@ -0,0 +1,36 @@
...
+    datetime
+)
+
+
+@collection(
+    name='publications',
+    unique_key='publication:identifier',
+    properties={
+        'title': 'Publications',
+        'description': 'Publication pages',
```
gabdank, line 18: I don't think the description should be "Publication pages", should not it be "Listing of publications"?

gabdank (changes requested):
please update the description of the collection of publications. Otherwise, I think it is good to go.

gabdank (approved):

PR #85, opened 2022-06-23
IGVF-203-organism-to-taxa

ian-whaling:


gabdank (changes requested):
I think you need to add update in changelogs of all the biosamples

gabdank (approved):

gabdank (approved):

PR #88, opened 2022-06-27
IGVF-208-date-obtained-sample

ian-whaling:
…nd inherited_sample

### DIFF
--- a/src/igvfd/schemas/changelogs/biosample.md
+++ b/src/igvfd/schemas/changelogs/biosample.md
@@ -4,3 +4,4 @@
 
 * Replaced related_biosamples and related_biosamples_relationship with pooled_from and part_of
 * organism property renamed to taxa
+* Bumped date_obtained property from biosample to sample.
--- a/src/igvfd/schemas/changelogs/biosample.md
+++ b/src/igvfd/schemas/changelogs/biosample.md
@@ -4,4 +4,4 @@
 
 * Replaced related_biosamples and related_biosamples_relationship with pooled_from and part_of
 * organism property renamed to taxa
-* Bumped date_obtained property from biosample to sample.
+* date_obtained property moved from Biosample schema to Sample schema. It is introduced in the Biosample schema through mixinProperties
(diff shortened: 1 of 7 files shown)


gabdank (changes requested):
please also add changelogs, in addition of the one comment

src/igvfd/schemas/changelogs/biosample.md
```diff
@@ -4,3 +4,4 @@
 
 * Replaced related_biosamples and related_biosamples_relationship with pooled_from and part_of
 * organism property renamed to taxa
+* Bumped date_obtained property from biosample to sample.
```
gabdank, line 7: I think we should use "Moved" instead of "Bumped"

gabdank: so probably something like: "*date_obtained* property definition was moved from the Biosample schema to Sample schema. It is introduced in Biosample through the mixin mechanism"

src/igvfd/schemas/changelogs/sample.md
```diff
@@ -0,0 +1,5 @@
+## Changelog for biosample.json
+
+### Minor changes since version 1
+
+* Bumped date_obtained property from biosample to sample.
```
gabdank, line 5: please see my comment on the changelog comment for Biosample, apply similar logic here.

src/igvfd/schemas/changelogs/technical_sample.md
```diff
@@ -3,3 +3,4 @@
 ### Changes since schema version 1
 
 * technical_sample has aliases mixin and submitted mixin removed, but still has properties for aliases and submitted due to change of basic_item mixin.  The actual properties have not changed.
+* Replaced date property with date_obtained which is inherited from sample.
```
gabdank, line 6: Even though we talk about inheritance, it is just mixing in properties from other schemas, so I would suggest avoiding saying inherited, instead, you can say "It is introduced in Technical sample schema through the mixin mechanism from Sample schema"

gabdank (changes requested):

gabdank (approved):

PR #90, opened 2022-06-28
IGVF-195-ancestors-ont

jenjou:


src/igvfd/types/term.py
```diff
@@ -59,6 +59,16 @@ def _get_ontology_slims(registry, term_id, slim_key):
...
 
+    @calculated_property(condition='term_id', schema={
+        'title': 'Ancestors',
+        'type': 'array',
+        'items': {
+            'type': 'string',
+        },
+    })
+    def ancestors(self, registry, term_id):
+        return self._get_ontology_slims(registry, term_id, 'ancestors')
```
gabdank, line 70: it is unclear what will be in ancestors' calculated list. perhaps a comment explaining what is going to be there is in order? explaining that these are upstream ontological terms, and then mentioning what is going to be in the list (IDs? term_names?..)

gabdank (changes requested):

src/igvfd/types/term.py
jenjou: Added a description for this property

gabdank (dismissed):

PR #96, opened 2022-07-01
IGVF-223 change-logs

gabdank:


src/igvfd/schemas/changelogs/human_donor.md
```diff
@@ -1,10 +1,12 @@
...
+### Minor changes since schema version 1
 
-* *taxon_id* now restricted to NCBI taxonomy ids that start with NCBI:txid followed by numbers. Example NCBI:txid9606 for *Homo sapiens*
-* *taxon_id* replaced by organism
-* defined *sex* as required property and defaulted it to *unspecified*
-* human_donor has aliases mixin and submitted mixin removed, but still has properties for aliases and submitted due to change of basic_item mixin.  The actual properties have not changed.
-* organism property renamed to taxa
-* Human donors now have a property *traits* to specify known phenotypic traits
+* Add `traits`.
+* Rename `organism` to `taxa`.
```
jenjou, line 6: Will we add the newest changes at the top or bottom of the lists? Just wondering if it is confusing here (since we happened to rename the same property twice). If we are consistent about adding to the top of the list, then this seems OK.

src/igvfd/schemas/changelogs/mixins.md
```diff
@@ -1,12 +1,10 @@
...
-
-* 3/24/22 Added product_id, lot_id, source mixins
-* 4/5/22 Added additional_description mixin
-* 4/7/22 Added product_id, lot_id, source mixins to a set called product_info
-* 4/27/22 Added collections
-* changed the property *date_created* into *creation_timestamp* and restricted the format to be date-time
-* 6/8/22 Added organism
-* removed mixins for aliases and submitted.  All properies for aliases and submitted have been moved under the basic_item mixin.  All schema objects that include the basic_item mixin will also have the properties previously in aliases and submitted.
-* 6/23/22 Renamed organism to taxa
+* Rename `organism` to `taxa`. (06/23/2022)
```
jenjou, line 3: Is it OK that we are including dates in the mixins changelog but not the other changelogs?

gabdank: I think yes, the dates on the mixins are replacing versioning, that is not there as it is not a schema

src/igvfd/schemas/changelogs/human_donor.md
gabdank: yes I have tried consistently to put latest changes on the top

jenjou (dismissed):

PR #98, opened 2022-07-05
IGVF-224 ontology term

gabdank:


jenjou (changes requested):
The code changes look correct, however, I think the existing "term.md" file is not yet removed from the branch and the entries in term.md could be added to the new ontology_term.md file.

src/igvfd/types/ontology_term.py
```diff
@@ -11,19 +11,19 @@
 
 
 @abstract_collection(
-    name='terms',
+    name='ontology_terms',
     unique_key='term:name',
```
jenjou, line 15: Should this line also be updated?

jenjou (approved):

jenjou:
Coverage test is failing but it seems to be flaky. We could try rerunning it.

PR #103, opened 2022-07-08
IGVF-207-biosample-requirements

khzlin:


src/igvfd/schemas/changelogs/biosample.md
```diff
@@ -1,5 +1,9 @@
 ## Changelog for *`biosample.json`*
 
+### Schema version 2
+
+* Added `donors` and `taxa` to requirements.
```
jenjou, line 5: We have decided to use present tense in the changelogs, so we should change "added" to "add" for consistency. This will affect all the other changelogs in this PR as well.

jenjou (changes requested):
I think the code looks good, just requesting small changes in the changelogs.

PR #106, opened 2022-07-13
IGVF-231-slim-image

ottojolanki:


docker/pyramid/Dockerfile
```diff
@@ -10,9 +10,19 @@ ENV REFERENCE_DATABASE_DIRECTORY=/srv/sqlite
...
+    make \
+    gcc \
+    git \
+    curl \
+    graphviz \
+    libpq-dev \
+    zlib1g-dev \
+    libjpeg-dev \
+    libmagic-dev \
+    libsqlite3-dev \
```
keenangraham, line 23: Do we want to order these?

keenangraham (approved):
I think fine overall. Probably worth the manual lib management for smaller size. Suppose can always go back to full image if it becomes too much of a pain.

docker/pyramid/Dockerfile
ottojolanki: yea, that makes sense.

PR #113, opened 2022-07-15
IGVF-230-use-shish-kebab-ontology-terms

meenakshikagda:


keenangraham (approved):
This looks fine, going to push through.

PR #117, opened 2022-07-20
IGVF-238-samples-descriptions

ian-whaling:


jenjou (changes requested):
I think we have been including the mixins changes in all schemas that inherit from it, so we should add an update in whole_organism.md as well.

I think this ticket might also conflict with IGVF-264, but we could probably discuss that. It might not be too much of a concern.

jenjou (dismissed):
Since there is no upgrade for whole_organism, there's no need for a changelog update. Code looks OK to proceed.

jenjou (commented):
Checked the schemas again and Award, Source, Document, and Whole Organism should cover all the cases with a separate (non mixin) `description` property. These 4 don't need a changelog note.

There are many schemas which are going to gain `description` via basic item, so we probably do need a changelog note for those. No upgrade needed.

jenjou (approved):

PR #122, opened 2022-07-26
IGVF-250-fixture-naming

ian-whaling:


src/igvfd/tests/fixtures/schemas/treatment.py
```diff
@@ -2,7 +2,7 @@
 
 
 @pytest.fixture
-def treatment_1(testapp):
+def treatment(testapp):
```
meenakshikagda, line 5: How about we make it treatment_chemical to match treatment_protein?

meenakshikagda: Not sure longer string but more description is better treatment_lactate_1_hr? @jenjou do you think this is acceptable?

jenjou: I like `treatment_chemical`. I'm having trouble thinking of cases where we want to specifically test the treatment duration, so I think that might be too much detail to include in the fixture name.

meenakshikagda: Okay

meenakshikagda:
Looks good

meenakshikagda (approved):

PR #129, opened 2022-08-09
IGVF-249-age-rework

ian-whaling:


src/igvfd/tests/data/inserts/differentiated_cell.json
```diff
@@ -19,7 +19,8 @@
             "3280b708-0f05-40fa-bff2-54e2b6505c13",
             "10c05ac0-52a2-11e6-bdf4-0800200c9a66"
         ],
-        "age": "90 or above",
+        "lower_bound_age": 90,
+        "upper_bound_age": 90,
```
cricketsloan, line 23: If the lower is 90 and the upper is 90, then isn't the age=90?

src/igvfd/tests/data/inserts/primary_cell.json
```diff
@@ -25,9 +25,9 @@
             "GEO:SAMN45"
         ],
         "taxa": "Homo sapiens",
-        "age": "40",
+        "lower_bound_age": 40,
+        "upper_bound_age": 45,
```
cricketsloan, line 29: If the age was 40 then shouldn't the lower=upper=40?

src/igvfd/tests/test_schema_cell_line.py
```diff
@@ -16,74 +16,54 @@ def test_lot_id_dependency(cell_line, testapp):
...
-        cell_line['@id'],
-        {'taxa': 'Mus musculus', 'life_stage': 'stationary'}, expect_errors=True)
-    assert res.status_code == 422
-    res = testapp.patch_json(
-        cell_line['@id'],
-        {'taxa': 'Saccharomyces', 'life_stage': 'adult'}, expect_errors=True)
-    assert res.status_code == 422
+
+
+# def test_lifestage_dependency(cell_line, testapp):
```
cricketsloan, line 54: There is commented code here

src/igvfd/tests/test_upgrade_cell_line.py
```diff
@@ -20,3 +20,18 @@ def test_cell_line_upgrade_3_4(upgrader, cell_line_v3):
...
+    assert value['lower_bound_age'] == 10 and value['upper_bound_age'] == 10
+    assert 'life_stage' not in value
+    assert value['schema_version'] == '5'
+    value = upgrader.upgrade('cell_line', cell_line_v4_unknown, current_version='4', target_version='5')
+    assert 'life_stage' not in value
+    assert 'age' not in value
+    assert value['schema_version'] == '5'
+    value = upgrader.upgrade('cell_line', cell_line_v4_90_or_above, current_version='4', target_version='5')
+    assert 'life_stage' not in value
+    assert value['lower_bound_age'] == 90 and value['upper_bound_age'] == 90
```
cricketsloan, line 36: I asked @gabdank this as well.  i am confused about how we are going to handle 90 and above now.  Is it 90 to 120?

src/igvfd/schemas/biosample.json
```diff
@@ -141,79 +112,24 @@
...
+            "title": "Lower Bound Age",
+            "description": "Lower bound of age of the organism at the time of collection of the sample.",
+            "comment": "For samples from human donors older than 89 years, the lower bound of age has to be 90 and upper bound has to be 90, for de-identification purposes.",
+            "type": "number"
+        },
+        "upper_bound_age": {
+            "title": "Upper Bound Age",
+            "description": "Upper bound of age of the organism at the time of collection of the sample.",
+            "comment": "For samples from human donors older than 89 years, the lower bound of age has to be 90 and upper bound has to be 90, for de-identification purposes.",
+            "type": "number"
```
cricketsloan, line 132: Can we use max value here to limit to 90?  Or likely in the dependencies to limit to 90 for human.

src/igvfd/schemas/changelogs/tissue.md
```diff
@@ -1,5 +1,12 @@
 ## Changelog for *`tissue.json`*
 
+### Schema version 5
+
+* Remove `life_stage`.
+* Convert `age` to be calculated and not submittable.
```
cricketsloan, line 6: To be calculated from lower_bound_age and upper_bound_age.

src/igvfd/tests/data/inserts/differentiated_cell.json
cricketsloan: NVM, I re-read the instructions for how this is supposed to work

src/igvfd/tests/data/inserts/primary_cell.json
cricketsloan: Maybe you are just trying to add diversity into the sample data?

src/igvfd/tests/test_upgrade_cell_line.py
cricketsloan: NVM, I have re-read the instructions again and I understand.  I am not sure we have a clean way of enforcing this plan.

cricketsloan (changes requested):
@gabdank I am concerned that if we remove lifestage and have a human sample that is 4 months that we will have no way of knowing if that is 4 months pre-natal or post-natal.

src/igvfd/tests/data/inserts/primary_cell.json
ian-whaling: Yes, that was my intention.

src/igvfd/schemas/biosample.json
ian-whaling: I don't think a max would work here as this maximum only applies to ages in age_units years.

ian-whaling: For example, a biosample can be 100 months old but not 100 years old.

cricketsloan: Could we add it in a dependency?  Maybe not

cricketsloan (approved):
Hey, I want to point out that one upgrader is used for all subclasses, and one age calculator is used for all subclasses.  So I think you can reduce the amount of tests.  I do like that there is some redundancy, it keeps you from say forgetting one subclass.  But I think we might have some overkill here.  

However, there is no reason to no approve for over testing.

PR #133, opened 2022-08-11
IGVF-239-a-post-diff-time-unit-update

KAndreeva:
IGVF-239 was redone to resolve conflicts.

### DIFF
--- a/src/igvfd/tests/test_schema_differentiated_cell.py
+++ b/src/igvfd/tests/test_schema_differentiated_cell.py
@@ -54,3 +54,30 @@ def test_part_of_differentiated_cell(differentiated_cell, differentiated_cell_pa
         differentiated_cell['@id'],
         {'part_of': differentiated_cell_part_of['@id']})
     assert res.status_code == 200
+
+
+def test_differentiation_time(differentiated_cell, testapp):
+    res = testapp.patch_json(
+        differentiated_cell['@id'],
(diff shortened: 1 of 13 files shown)


src/igvfd/tests/test_schema_differentiated_cell.py
```diff
@@ -54,3 +54,30 @@ def test_part_of_differentiated_cell(differentiated_cell, differentiated_cell_pa
...
+    assert res.status_code == 422
+
+    res = testapp.patch_json(
+        differentiated_cell['@id'],
+        {'post_differentiation_time_units': 'hour'}, expect_errors=True)
+    assert res.status_code == 422
+
+    res = testapp.patch_json(
+        differentiated_cell['@id'],
+        {'post_differentiation_time': 10, 'post_differentiation_time_units': 'hour'}, expect_errors=True)
```
cricketsloan, line 72: expect_errors=True, but i don't think you expect errors here

cricketsloan (approved):
I think in general, when we are not expected in errors we do not put expect_errors=True, but I don't think it breaks if you do.

src/igvfd/tests/test_schema_differentiated_cell.py
KAndreeva: eliminated 'expect_errors=True'

KAndreeva:
Closing this branch and will have new branch for the rebase.

PR #139, opened 2022-08-22
IGVF-242-calculated-summary-property

jenjou:


gabdank (changes requested):
@jenjou I think you need to add some types tests. For some concrete classes. And check if replaced status behavior is handled correctly

gabdank (dismissed):

gabdank (approved):

PR #147, opened 2022-08-31
IGVF-290-add-revoke-detail-property

jenjou:


gabdank (changes requested):
please add revoked insert with revoke_detail

gabdank (approved):

PR #152, opened 2022-09-06
IGVF-155-file

jenjou:


jenjou:
@cricketsloan This ticket isn't "complete" yet, but I thought it would be helpful to have an initial review of what we have so far. We are missing a subschema that could be used for processed/non-raw data. Also, I need to remove the unused schema "data.json".

src/igvfd/schemas/file.json
```diff
@@ -0,0 +1,151 @@
...
+            "description": "The type of content in the file.",
+            "comment": "This property is overwritten by the subclasses to define specific enum values.",
+            "type": "string",
+            "enum": []
+        },
+        "dbxrefs": {
+            "@type": "@id",
+            "rdfs:subPropertyOf": "rdfs:seeAlso",
+            "title": "External Resources",
+            "description": "Identifiers from external resources that may have 1-to-1 or 1-to-many relationships with IGVF objects.",
```
gabdank, line 71: I think it should be specific to "IGVF file objects" rather than general "IGVF objects"

src/igvfd/schemas/file.json
```diff
@@ -0,0 +1,151 @@
...
+            "comment": "This property is overwritten by the subclasses to define specific enum values.",
+            "type": "array",
+            "uniqueItems": true,
+            "items": {
+                "title": "External identifier",
+                "description": "Identifier from an external resource that may have 1-to-1 or 1-to-many relationships with IGVF objects.",
```
gabdank, line 77: same comments as above "IGVF file objects"

src/igvfd/schemas/sequence_data.json
```diff
@@ -0,0 +1,71 @@
+{
+    "title": "Sequence data",
+    "$id": "/profiles/sequence_data.json",
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "required": [
+        "award",
+        "lab"
```
gabdank, line 7: should not the required properties list be expanded in the sequence_data?

src/igvfd/schemas/sequence_data.json
```diff
@@ -0,0 +1,71 @@
...
+        "$merge": [
+            "file.json#/dependentSchemas"
+        ]
+    },
+    "type": "object",
+    "properties": {
+        "schema_version": {
+            "default": "1"
+        },
+        "content_type": {
```
gabdank, line 42: I know that it may be very long, but perhaps we should add a comment with an explanation of what each of the enums is? with examples?

src/igvfd/types/file.py
```diff
@@ -0,0 +1,36 @@
...
+@collection(
+    name='sequence-data',
+    unique_key='accession',
+    properties={
+        'title': 'Sequence data',
+        'description': 'Listing of sequence data files',
+    })
+class SequenceData(File):
+    item_type = 'sequence_data'
+    schema = load_schema('igvfd:schemas/sequence_data.json')
```
gabdank, line 36: I think we need to add the md5 uniqueness function from ENCODE for the sequence_data

gabdank (changes requested):

gabdank (changes requested):
i think we should uniformly use either "File" or "file" otherwise it looks like in some places it is capitalized and in other it is not, while I don't think we mean there is difference between the object we are talking about

src/igvfd/schemas/imaging_data.json
```diff
@@ -0,0 +1,96 @@
...
+        },
+        "content_type": {
+            "comment": "Content Type describes the content of the file.",
+            "enum": [
+                "image"
+            ]
+        },
+        "file_format": {
+            "enum": [
+                "tif"
```
gabdank, line 57: I am wondering if, instead of TIF we would expect to get TAR-ball with multiple TIFFs?

gabdank (changes requested):

src/igvfd/schemas/sequence_data.json
```diff
@@ -0,0 +1,132 @@
...
+                "title": "External identifier",
+                "description": "Identifier from an external resource that may have 1-to-1 or 1-to-many relationships with IGVF file objects.",
+                "type": "string",
+                "pattern": "^(SRA:(SRR|SRX)\\d+)$"
+            }
+        },
+        "platform": {
+            "title": "Platform",
+            "description": "The sequencing platform used to produce the sequence data.",
+            "type": "string"
```
gabdank, line 90: I think platform is required for FASTQ

src/igvfd/schemas/sequence_data.json
```diff
@@ -0,0 +1,132 @@
...
+            "minimum": 0
+        },
+        "maximum_read_length": {
+            "title": "Read Length",
+            "description": "For high-throughput sequencing, the maximum number of contiguous nucleotides determined by sequencing.",
+            "type": "integer",
+            "minimum": 0
+        },
+        "mean_read_length": {
+            "title": "Read Length",
```
gabdank, line 105: I think read length, as well as reads count, should be calculated and fille din by the validator. So they should be required - after the uploading status changed to validated. not sure it is easy to do in dependencies

gabdank (changes requested):

gabdank (approved):

gabdank (dismissed):

gabdank (approved):

PR #156, opened 2022-09-19
IGVF-299-CuratedSet

ian-whaling:


src/igvfd/schemas/changelogs/curated_set.md
```diff
@@ -0,0 +1,3 @@
+## Changelog for *`curated_set.json`*
+
+### Minor changes since schema version 1
```
gabdank, line 3: I don't think there are changes so the line about minor changes shold not be there?

src/igvfd/schemas/curated_set.json
```diff
@@ -0,0 +1,61 @@
...
+        },
+        {
+            "$ref": "mixins.json#/taxa"
+        }
+    ],
+    "properties": {
+        "schema_version": {
+            "default": "1"
+        },
+        "reference_type": {
```
gabdank, line 47: after thinking about it, I think we should not call it "reference_type" but "curated_set_type"

gabdank (changes requested):

gabdank (approved):

gabdank (approved):

PR #166, opened 2022-10-06
IGVF-320 Updated not-submittable true for all calculated properties

KAndreeva:
Updated not submittable true for all calculated properties


jenjou (approved):

gabdank (changes requested):
Kalina, I think you can drop the changelog updates. If you feel strongly that the changelog should be updated, we can leave it, but certainly shorten it, by removing: the "properties" or "calculated properties" suffix in the sentences. The change log is to be brief and consice.

KAndreeva:
Closing pull request for a new branch with a shorter name.

PR #167, opened 2022-10-07
IGVF-298-analysis-set

jenjou:


src/igvfd/schemas/analysis_set.json
```diff
@@ -0,0 +1,53 @@
+{
+    "title": "Analysis Set",
+    "$id": "/profiles/analysis_set.json",
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "description": "Abstract schema for an analysis set.",
```
gabdank, line 5: I don't think this an abstract class

gabdank (changes requested):

gabdank (dismissed):

src/igvfd/types/file_set.py
```diff
@@ -22,3 +22,39 @@ class FileSet(Item):
...
+    )
+    def assay_title(self, request, input_file_set=None):
+        assay_title = []
+        if input_file_set is not None:
+            for fileset in input_file_set:
+                file_set_object = request.embed(fileset, '@@object')
+                if file_set_object.get('assay_title') and \
+                        'MeasurementSet' in file_set_object.get('@type'):
+                    assay_title.append(file_set_object.get('assay_title'))
+            return list(set(assay_title))
```
gabdank, line 60: I think you can define assay_title as "set()", instead of appending you can use the method "add()" and then in the end you don't need to convert list into set and back into list, instead just convert set into list

gabdank (changes requested):
minor suggestion about lists and sets

gabdank (approved):

PR #171, opened 2022-10-25
IGVF-332 Added placeholder for CCF uuid

KAndreeva:
Added placeholder for CCF uuid.

### DIFF
--- a/src/igvfd/schemas/tissue.json
+++ b/src/igvfd/schemas/tissue.json
@@ -51,7 +51,18 @@
     "dependentSchemas": {
         "$merge": [
             "biosample.json#/dependentSchemas"
-        ]
+        ],
+        "ccf_id": {
+            "comment": "Common coordinate framework identifier only applies to human tissue samples.",
+            "required": [
+                "taxa"
+            ],
+            "properties": {
+                "taxa": {
+                    "const": "Homo sapiens"
+                }
+            }
+        }
     },
     "properties": {
         "schema_version": {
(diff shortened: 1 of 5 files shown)


src/igvfd/schemas/tissue.json
```diff
@@ -77,6 +88,14 @@
                 "week"
             ]
         },
+        "ccf_id": {
+            "title": "Common coordinate framework identifier",
+            "description": "This identifier associates an entry from the HubMap Consortium for specimens, biological structures, and spatial positions.",
```
gabdank, line 93: can we please change description to : "HubMap Common Coordinate Framework unique identifier corresponding to the organ, biological structure, and spatial location of the tissue specimen."

src/igvfd/schemas/tissue.json
```diff
@@ -77,6 +88,14 @@
...
+            "comment": "This identifier will be later linked to: https://portal.hubmapconsortium.org/ccf-eui",
+            "type": "string",
+            "format": "uuid",
+            "nullable": true
```
gabdank, line 97: why should it be "nullable" true?

gabdank: either we have or we don't, i don't think we need nullable

src/igvfd/tests/data/inserts/tissue.json
```diff
@@ -110,6 +112,7 @@
         "part_of": "IGVFSM000KAS",
         "source": "/labs/j-michael-cherry/",
         "nih_institutional_certification": "NICHD1442",
+        "ccf_id": null,
```
gabdank, line 115: i don't think we need this

gabdank (changes requested):

src/igvfd/schemas/tissue.json
KAndreeva: Done

src/igvfd/schemas/tissue.json
KAndreeva: done

gabdank (approved):

PR #179, opened 2022-11-18
IGVF-325-construct-library-schema

iyoungworth:


src/igvfd/schemas/construct_library.json
```diff
@@ -0,0 +1,317 @@
...
+        },
+        "scope": {
+            "title": "Library Construct Scope",
+            "description": "The scope or scale that this construct library is designed to target. If the scope is across gene(s) or chromosomal region(s), these will need to be specified in the targeted_genes or targeted_regions property.",
+            "type": "string",
+            "enum": [
+                "genes",
+                "chromosomal regions",
+                "genome-wide",
+                "population"
```
gabdank, line 109: not sure about population? if we add it, then it should be explicitly explained in comment

src/igvfd/schemas/construct_library.json
```diff
@@ -0,0 +1,317 @@
...
+            "minItems": 1,
+            "uniqueItems": true,
+            "items": {
+                "comment": "See gene.json for a list of available identifiers.",
+                "type": "string",
+                "linkTo": "Gene"
+            }
+        },
+        "targeted_regions": {
+            "title": "Targeted Loci",
```
gabdank, line 125: I think we want singular, not plural for property names

src/igvfd/schemas/construct_library.json
```diff
@@ -0,0 +1,317 @@
...
+        "accession": {
+            "accessionType": "FS"
+        },
+        "scope": {
+            "title": "Library Construct Scope",
+            "description": "The scope or scale that this construct library is designed to target. If the scope is across gene(s) or chromosomal region(s), these will need to be specified in the targeted_genes or targeted_regions property.",
+            "type": "string",
+            "enum": [
+                "genes",
+                "chromosomal regions",
```
gabdank, line 107: maybe "locus" instead of "chromosomal region"?

src/igvfd/schemas/construct_library.json
```diff
@@ -0,0 +1,317 @@
...
+            "type": "array",
+            "minItems": 1,
+            "uniqueItems": true,
+            "items": {
+                "comment": "See gene.json for a list of available identifiers.",
+                "type": "string",
+                "linkTo": "Gene"
+            }
+        },
+        "targeted_regions": {
```
gabdank, line 124: region -> locus

src/igvfd/schemas/construct_library.json
iyoungworth: Removed for now, we can add back if given a definition by an IGVF lab

src/igvfd/schemas/construct_library.json
iyoungworth: Understood, all are singularized

src/igvfd/schemas/construct_library.json
```diff
@@ -0,0 +1,378 @@
...
+        },
+        "mpra_library_detail": {
+            "title": "MPRA Library Detail"
+        }
+    },
+    "boost_values": {
+        "accession": 1.0,
+        "aliases": 1.0,
+        "@type": 1.0,
+        "targeted_gene": 1.0
```
iyoungworth, line 375: As far as I can tell, boost_values for nested properties is not accepted (as it was in encoded). I tried targeted_gene.genes.symbol and other variations but all resulted in errors in cdk-pytest

src/igvfd/schemas/construct_library.json
```diff
@@ -0,0 +1,378 @@
...
+                        "description": "The starting coordinate.",
+                        "type": "integer",
+                        "minimum": 0
+                    },
+                    "end": {
+                        "title": "End",
+                        "description": "The ending coordinate.",
+                        "type": "integer",
+                        "minimum": 0
+                    }
```
gabdank, line 202: I think we need to explicitly say it is 1-based

src/igvfd/schemas/construct_library.json
```diff
@@ -0,0 +1,378 @@
...
+            "title": "Average Size",
+            "description": "The average size of the inserts (such as guides) cloned into vectors in the library.",
+            "comment": "May not be submitted with size_range.",
+            "type": "number",
+            "minimum": 0
+        },
+        "size_range": {
+            "title": "Size Range",
+            "description": "The numerical range of the size of the inserts (such as guides) cloned in vectors in the library..",
+            "comment": "May not be submitted with average_size.",
```
gabdank, line 228: remove extra "." in the description

src/igvfd/schemas/construct_library.json
gabdank: OK

gabdank (changes requested):
few minor changes

src/igvfd/schemas/construct_library.json
iyoungworth: I mentioned it in the overall 'targeted_locus' comment, but I've added "1-based" to each of these sub-properties comments' as well

src/igvfd/schemas/construct_library.json
```diff
@@ -0,0 +1,378 @@
...
+            "items": {
+                "comment": "See gene.json for a list of available identifiers.",
+                "type": "string",
+                "linkTo": "Gene"
+            }
+        },
+        "targeted_locus": {
+            "title": "Targeted Locus",
+            "description": "The specific chromosomal region(s) this construct library was designed to target.",
+            "comment": "Each locus is defined using 1-based, closed coordinates for a specific version of the reference genome assembly.",
```
gabdank, line 161: I am not sure what "closed" coordinates mean

src/igvfd/schemas/construct_library.json
```diff
@@ -0,0 +1,378 @@
...
+                        "enum": [
+                            "GRCh38",
+                            "GRCm39"
+                        ]
+                    },
+                    "chromosome": {
+                        "title": "Chromosome",
+                        "description": "The number (or letter) designation for the chromosome, e.g. chr1 or chrX",
+                        "type": "string",
+                        "pattern": "^(chr[0-9A-Za-z]+)$"
```
gabdank, line 189: would chrY work?

src/igvfd/schemas/construct_library.json
```diff
@@ -0,0 +1,378 @@
...
+                        "title": "Start",
+                        "description": "The 1-based starting coordinate.",
+                        "type": "integer",
+                        "minimum": 0
+                    },
+                    "end": {
+                        "title": "End",
+                        "description": "The 1-based ending coordinate.",
+                        "type": "integer",
+                        "minimum": 0
```
gabdank, line 201: you have to be specific if it is included or not, I think Cricket had some proper terminology for these type of coordinates specifications

src/igvfd/schemas/construct_library.json
```diff
@@ -0,0 +1,378 @@
...
+        "average_size": {
+            "title": "Average Size",
+            "description": "The average size of the inserts (such as guides) cloned into vectors in the library.",
+            "comment": "May not be submitted with size_range.",
+            "type": "number",
+            "minimum": 0
+        },
+        "size_range": {
+            "title": "Size Range",
+            "description": "The numerical range of the size of the inserts (such as guides) cloned in vectors in the library.",
```
gabdank, line 227: in general, I think examples (such as) belong in comments. And I am a bit confused about the meaning of size_range for MPRA vs. CRISPR guides.

src/igvfd/schemas/construct_library.json
```diff
@@ -0,0 +1,378 @@
...
+                "title": "MPRA Library",
+                "type": "object",
+                "additionalProperties": false,
+                "properties": {
+                    "complexity": {
+                        "title": "Complexity",
+                        "description": "The complexity (number of constructs) in the library.",
+                        "type": "integer",
+                        "minimum": 0
+                    }
```
gabdank, line 248: aren't you already asking for quantity? how is that different from complexity?

src/igvfd/schemas/construct_library.json
iyoungworth: Yes, X and Y are accepted

src/igvfd/schemas/construct_library.json
iyoungworth: Described in detail on the ticket, now broken to mpra_library_complexity and average_guide_coverage/guide_coverage_range

src/igvfd/schemas/construct_library.json
iyoungworth: Understood, removed. The language says the size property should describe the insert, which I think should apply to either MPRA or CRISPR guides (and the property is allowed for either)

src/igvfd/schemas/construct_library.json
```diff
@@ -0,0 +1,368 @@
...
+            },
+            "comment": "Specification of size_range is mutually exclusive with specification of average_size."
+        }
+    },
+    "properties": {
+        "schema_version": {
+            "default": "1"
+        },
+        "accession": {
+            "accessionType": "FS"
```
gabdank, line 118: there is Jennifers' ticket that will change it to be DS instead of FS
probably need to wait for that one to get merged? or the other way around

src/igvfd/schemas/construct_library.json
```diff
@@ -0,0 +1,368 @@
...
+            "title": "Library Construct Scope",
+            "description": "The scope or scale that this construct library is designed to target. If the scope is across gene(s) or loci, these will need to be specified in the targeted_genes or targeted_loci property.",
+            "type": "string",
+            "enum": [
+                "genes",
+                "loci",
+                "genome-wide"
+            ]
+        },
+        "origin": {
```
gabdank, line 130: I think there could be multiple origins? like library from histone and TF or TF and DNase. I think should be a plural and a list

src/igvfd/schemas/construct_library.json
```diff
@@ -0,0 +1,368 @@
...
+            "minimum": 0
+        },
+        "size_range": {
+            "title": "Size Range",
+            "description": "The numerical range of the size of the inserts cloned in vectors in the library.",
+            "comment": "May not be submitted with average_size.",
+            "type": "string",
+            "pattern": "(^[0-9]+-[0-9]+$|^[<>][0-9]+$)"
+        },
+        "mpra_library_details": {
```
gabdank, line 231: should not this be just a property "MPRA library complexity" ? rather than object with this single property?

gabdank: unless we have some additional properties in mind it is hard to justify object? maybe directionality of inserts? or location of element in respect to the reporter gene? maybe reporter gene?

src/igvfd/schemas/construct_library.json
```diff
@@ -0,0 +1,368 @@
...
+                    ]
+                },
+                "average_guide_coverage": {
+                    "title": "Average Guide Coverage",
+                    "description": "The average number of guides targeting each element of interest in the library.",
+                    "comment": "May not be submitted with guide_coverage_range.",
+                    "type": "number",
+                    "minimum": 0
+                },
+                "guide_coverage_range": {
```
gabdank, line 299: I suggest to convert into min and max in the range, otherwise it is not numeric and is difficult to interpret/use

gabdank (changes requested):

src/igvfd/schemas/construct_library.json
```diff
@@ -0,0 +1,368 @@
...
+        "uuid",
+        "aliases",
+        "accession",
+        "alternate_accessions"
+    ],
+    "additionalProperties": false,
+    "mixinProperties": [
+        {
+            "$ref": "mixins.json#/basic_item"
+        },
```
gabdank, line 22: are we not mixing in any FileSet properties?

gabdank (changes requested):

src/igvfd/schemas/construct_library.json
iyoungworth: Changed the phrasing to what we discussed (The 1-based, closed (inclusive) ending coordinate.)

src/igvfd/schemas/construct_library.json
iyoungworth: Have now removed this as discussed (lack a consistent definition between different labs doing MPRA)

src/igvfd/schemas/construct_library.json
iyoungworth: Rebased

src/igvfd/schemas/construct_library.json
iyoungworth: As discussed, size of fragments (lower_ and upper_bound_insert_size, average_insert_size) are now in this sub-object. We expect to add something like reporter gene, directionality, other details of elements in the vector if labs want these details to be searchable/obvious on the page

src/igvfd/schemas/construct_library.json
iyoungworth: Done, did this with fragment size as well

src/igvfd/schemas/construct_library.json
iyoungworth: Currently FileSet only has unique properties (non-mixins) Donors and Samples- do we want either of these?

src/igvfd/schemas/measurement_set.json
```diff
@@ -54,6 +54,20 @@
             "description": "Link to the protocol for conducting the assay on Protocols.io.",
             "type": "string",
             "pattern": "^https://www\\.protocols\\.io/(\\S*)$"
+        },
+        "construct_library": {
```
gabdank, line 58: I think this one has to be plural? array of relevant construct libraries?

src/igvfd/schemas/measurement_set.json
```diff
@@ -54,6 +54,20 @@
...
+            "title": "Construct Library",
+            "description": "The construct library of vectors introduced to the biosample used to produce data in this measurement set.",
+            "comment": "See construct_library.json for available identifiers.",
+            "type": "array",
+            "minItems": 1,
+            "maxItems": 1,
```
gabdank, line 64: not sure why it has maxItems here?

gabdank (changes requested):
plural vs singular and maxItem questions

src/igvfd/schemas/measurement_set.json
iyoungworth: Makes sense, adjusted to be plural- we talked about restricting to 1 ConstructLibrary (similar to the logic for 1 SequencingLibrary, 1 Biosample per MeasurementSet) which is why maxItems is 1 (see below comment), but we would keep it an array in case anyone needed to submit multiple. I'm fine with removing maxItems as I don't think it would necessarily create confusion to have multiple

src/igvfd/schemas/measurement_set.json
iyoungworth: See above for detailed note

gabdank (approved):

src/igvfd/schemas/construct_library.json
```diff
@@ -0,0 +1,353 @@
...
+            "additionalProperties": false,
+            "required": [
+                "guide_type"
+            ],
+            "dependentSchemas": {
+                "average_guide_coverage": {
+                    "not": {
+                        "required": [
+                            "upper_bound_coverage_range",
+                            "lower_bound_coverage_range"
```
jenjou, line 278: These property names don't match up with the property names below (lines 333-338) e.g. "lower_bound_guide_coverage".

src/igvfd/schemas/construct_library.json
```diff
@@ -0,0 +1,353 @@
...
+                    ]
+                },
+                "average_guide_coverage": {
+                    "title": "Average Guide Coverage",
+                    "description": "The average number of guides targeting each element of interest in the library.",
+                    "type": "number",
+                    "minimum": 0
+                },
+                "lower_bound_guide_coverage": {
+                    "title": "Lower Bound Insert Size",
```
jenjou, line 334: Possible copy paste error, I think this title is matching the reporter_library_details subproperties and probably needs to be updated.

src/igvfd/schemas/construct_library.json
```diff
@@ -0,0 +1,353 @@
...
+                    "description": "Lower bound of the number of guides targeting each element of interest in the library.",
+                    "type": "number"
+                },
+                "upper_bound_guide_coverage": {
+                    "title": "Upper Bound Insert Size",
```
jenjou, line 339: Same as above, same name as the property in reporter_library_details

src/igvfd/schemas/construct_library.json
```diff
@@ -0,0 +1,353 @@
+{
+    "title": "Construct library",
```
jenjou, line 2: We should capitalize Library for consistency.

src/igvfd/schemas/construct_library.json
iyoungworth: Thank you for catching this- too many similarly-named properties. All are adjusted to upper_ and lower_bound_guide_coverage

src/igvfd/schemas/construct_library.json
```diff
@@ -0,0 +1,362 @@
...
+                        "title": "Start",
+                        "description": "The 1-based, closed (inclusive) starting coordinate.",
+                        "type": "integer",
+                        "minimum": 0
+                    },
+                    "end": {
+                        "title": "End",
+                        "description": "The 1-based, closed (inclusive) ending coordinate.",
+                        "type": "integer",
+                        "minimum": 0
```
gabdank, line 199: if the coordinates are 1-based should not the minimum be 1?

gabdank (changes requested):
one change of the position spec assuming we are using 1-based position

src/igvfd/schemas/construct_library.json
iyoungworth: Ah, yes- my mistake keeping that from ENCODE! I updated to be minimum=1 and used it in a test to confirm '1' was accepted

gabdank (dismissed):

gabdank (dismissed):

gabdank (approved):

PR #185, opened 2022-11-29
IGVF-353-phenotypic-feature

jenjou:


src/igvfd/schemas/phenotypic_feature.json
```diff
@@ -0,0 +1,88 @@
...
+        "aliases"
+    ],
+    "additionalProperties": false,
+    "mixinProperties": [
+        {
+            "$ref": "mixins.json#/basic_item"
+        },
+        {
+            "$ref": "mixins.json#/standard_status"
+        }
```
gabdank, line 21: if there is no attribution, labs would never be able to submit these? is that intended?

gabdank (changes requested):

src/igvfd/schemas/phenotypic_feature.json
jenjou: Not intended, I will add it in.

gabdank (approved):

gabdank (approved):

PR #188, opened 2022-12-13
IGVF-366-lab-aliases

ian-whaling:


jenjou (changes requested):
Small typo in one name, "honjun-song" should be "hongjun-song"

jenjou (dismissed):

PR #194, opened 2023-01-09
IGVF-321-singular-lists

ian-whaling:


src/igvfd/types/base.py
```diff
@@ -206,7 +206,7 @@ def __ac_local_roles__(self):
         if 'lab' in properties:
             lab_submitters = 'submits_for.%s' % properties['lab']
             roles[lab_submitters] = 'role.lab_submitter'
-        if 'award' in properties:
+        if 'award' in properties and type(properties['award']) != list:
```
ian-whaling, line 209: The award property in lab was changed from 'awards' to 'award,' so this '__ac_local_roles__ ' property would now be evaluated for lab and it requires that the 'award' property be a string, but in lab it is a list, so I added this line to avoid it with lab. If you prefer a different way I can change it.

keenangraham: The idiomatic way would be something like `isinstance(properties['award'], list)`, though I'm not sure this change in general is a good idea.

keenangraham (changes requested):
I have some concerns in general about changing all plural names to singular, when they are actually pointing to a collection of things.

Also will need deeper review since this is touching authentication/authorization things.

PR #199, opened 2023-01-12
IGVF-387 revert donors

gabdank:


jenjou (commented):
Do we need to update File Set schema (and related tests/inserts) in this ticket as well?

jenjou (approved):
We will not change File Set in this ticket, since this ticket is only to revert the previous change in the biosamples. File Set changes will go in a different ticket.

jenjou (approved):

PR #206, opened 2023-01-23
IGVF-368-filter-dbinstanceidentifier-in-python

ottojolanki:


cdk/infrastructure/runtime/lambdas/snapshot/main.py
```diff
@@ -47,6 +47,11 @@ def sort_results_by_create_time(results):
     )
 
 
+def filter_results_by_db_instance_identifier(results, db_instance_identifier):
+    filtered_results = [result for result in results if result['DBInstanceIdentifier'] == db_instance_identifier]
```
keenangraham, line 51: Everything will always have a `DBInstanceIdentifier`, right?

keenangraham (approved):
LGTM. We should try a couple different restore from snapshot scenarios on the demo.

cdk/infrastructure/runtime/lambdas/snapshot/main.py
ottojolanki: Yes, that should be the case according to boto docs.

keenangraham (dismissed):
I think this looks good.

Once the snaphot_share is deployed in production we should QA demo here that can:
1. Restore from latest dev database ID (like now).
2. Restore from latest prod database ID (new).
3. Restore from direct snapshot ID (local or shared).

keenangraham (approved):

PR #223, opened 2023-02-08
IGVF-308-object-summary

ian-whaling:


src/igvfd/types/sample.py
```diff
@@ -63,15 +63,15 @@ def sex(self, request, donors=None):
...
-    @calculated_property(
-        schema={
-            'title': 'Age',
-            'description': 'Age of organism at the time of collection of the sample.',
-            'type': 'string',
-            'pattern': '^((\\d+(\\.[1-9])?(\\-\\d+(\\.[1-9])?)?)|(unknown)|([1-8]?\\d)|(90 or above))$',
-            'notSubmittable': True,
-        }
-    )
+    @calculated_property(define=True,
```
gabdank, line 66: why here it is "define=True" but in others, it is not?

src/igvfd/types/sample.py
```diff
@@ -107,6 +122,21 @@ class InVitroSystem(Biosample):
...
 
+    @calculated_property(
+        schema={
+            'title': 'Summary',
+            'type': 'string',
+            'notSubmittable': True,
+        }
+    )
+    def summary(self, request, biosample_term, taxa, classification=None, time_post_factors_introduction=None, time_post_factors_introduction_units=None):
+        sample_term_object = request.embed(biosample_term, '@@object')
```
gabdank, line 133: i am wondering if you need calculated properties when you embed the underlying sample_term or not, I think there is a way to skip calculated properties making it more efficient

gabdank: that is true for all of the calculated properties and embedding in them

src/igvfd/types/sample.py
ian-whaling: `define=True` is necessary here in `age` to be able to pass it into another calculated property, i.e. `summary`

src/igvfd/types/sample.py
ian-whaling: skipped calculated properties with '@@object?skip_calculated=true'

src/igvfd/types/sample.py
```diff
@@ -107,6 +123,21 @@ class InVitroSystem(Biosample):
...
+            'title': 'Summary',
+            'type': 'string',
+            'notSubmittable': True,
+        }
+    )
+    def summary(self, request, biosample_term, taxa, classification=None, time_post_factors_introduction=None, time_post_factors_introduction_units=None):
+        sample_term_object = request.embed(biosample_term, '@@object?skip_calculated=true')
+        term_name = sample_term_object.get('term_name')
+        if time_post_factors_introduction and time_post_factors_introduction_units:
+            return f'{term_name} {classification}, {taxa} ({time_post_factors_introduction} {time_post_factors_introduction_units})'
```
jenjou, line 137: Wondering if we should add more detail here about the time post factors introduction. Without additional context, my first assumption would be that the time refers to age.

gabdank: I think we can skip the verbs at this time
if and when labs complain we can consider adding verbs

src/igvfd/tests/test_types_biosample.py
```diff
@@ -36,10 +36,94 @@ def test_age_calculation(testapp, in_vitro_cell_line):
...
-    print(res.json.get('age'))
     assert res.json.get('age') == '90 or above'
     res = testapp.patch_json(
         in_vitro_cell_line['@id'],
         {'age_units': 'month'})
     res = testapp.get(in_vitro_cell_line['@id'])
     assert res.json.get('age') == '90'
+
+
+def test_summary(testapp, tissue, primary_cell, whole_organism, in_vitro_cell_line, technical_sample, sample_term_endothelial_cell, sample_term_embryoid_body, sample_term_lymphoblastoid):
```
jenjou, line 47: Can we add a test here for the case where 'tissue' is in the term name and in the biosample type or classification?

src/igvfd/types/sample.py
```diff
@@ -107,6 +131,32 @@ class InVitroSystem(Biosample):
...
+        }
+    )
+    def summary(self, request, biosample_term, taxa, classification, time_post_factors_introduction=None, time_post_factors_introduction_units=None):
+        sample_term_object = request.embed(biosample_term, '@@object?skip_calculated=true')
+        term_name = sample_term_object.get('term_name')
+        term_and_classification = f'{term_name} {classification}'  # default case
+        if classification in term_name:
+            term_and_classification = term_name  # if classification in term_name, drop classification from summary
+        elif 'cell' in classification and 'cell' in term_name:
+            term_name = term_name.replace('cell', classification)
```
jenjou, line 148: Minor suggestion, but this can be combined with the following line instead of splitting it in two. (so "term_and_classification = term_name.replace(...". Likewise on line 150.

jenjou (changes requested):

src/igvfd/tests/test_types_biosample.py
```diff
@@ -36,10 +36,123 @@ def test_age_calculation(testapp, in_vitro_cell_line):
...
+        in_vitro_cell_line['@id'],
+        {
+            'biosample_term': sample_term_brown_adipose_tissue['@id'],
+            'classification': 'differentiated tissue',
+            'time_post_factors_introduction': 1,
+            'time_post_factors_introduction_units': 'month'
+        }
+    )
+    res = testapp.get(in_vitro_cell_line['@id'])
+    print(res.json.get('summary'))
```
jenjou, line 155: Extra line here, can be dropped

jenjou (dismissed):
Looks OK, just need to remove one extra line that was left in.

jenjou (changes requested):
Looks OK, just need to remove one extra line that was left in.

PR #231, opened 2023-02-16
IGVF-444-remove-traits-from-donor

KAndreeva:
Removed traits from donor, human donor and rodent donor. Added update code for new schema version.

### DIFF
--- a/src/igvfd/schemas/changelogs/donor.md
+++ b/src/igvfd/schemas/changelogs/donor.md
@@ -2,6 +2,7 @@
 
 ### Release v2
 
+* Remove `traits`.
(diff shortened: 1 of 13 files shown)


src/igvfd/schemas/changelogs/donor.md
```diff
@@ -1,5 +1,9 @@
 ## Changelog for *`donor.json`*
```
gabdank, line 2: I think we agreed not to put versions in abstract classes?

gabdank (dismissed):
changelog

gabdank (approved):

PR #240, opened 2023-02-28
IGVF-460-add-report-download

mingjiecn:


src/igvfd/report.py
```diff
@@ -0,0 +1,140 @@
...
+        downloadtime.month,
+        downloadtime.day,
+        downloadtime.hour,
+        downloadtime.minute
+    )
+    request.response.app_iter = generate_rows()
+    return request.response
+
+
+def list_visible_columns_for_schemas(request, schemas):
```
keenangraham, line 103: Should update this to look for columns in the search config for a given type.

src/igvfd/report.py
```diff
@@ -0,0 +1,140 @@
...
+    types = request.params.getall('type')
+    if len(types) != 1:
+        msg = 'Report view requires specifying a single type.'
+        raise HTTPBadRequest(explanation=msg)
+
+    # Make sure we get all results
+    request.GET['limit'] = 'all'
+    type_str = types[0]
+    schemas = [request.registry[TYPES][type_str].schema]
+    columns = list_visible_columns_for_schemas(request, schemas)
```
keenangraham, line 68: I think we need to update this to look for `columns` in the `search-config-registry` instead of in `schemas`, now that they have been moved out. Should be able to access the configs through the request (like here https://github.com/IGVF-DACC/igvfd/blob/b9e160feea5d603d750d7f2dba6d8934b76dbf87/src/igvfd/search_views.py#L259)

keenangraham (changes requested):
This looks good but I think we need to update the logic to look for `columns` in the search config for a given type, instead of in the schema. Otherwise the default columns in the report don't really make sense:
Screen Shot 2023-03-01 at 10 16 16 AM

keenangraham (approved):
LGTM.

keenangraham (approved):

PR #241, opened 2023-03-01
IGVF-476-reverse-link-file-sets

ian-whaling:


src/igvfd/tests/test_types_sample.py
```diff
@@ -0,0 +1,26 @@
+import pytest
+
+
+def test_summary(testapp, tissue, measurement_set, analysis_set_base, curated_set_genome):
```
gabdank: why is the test called "summary"?

src/igvfd/tests/test_types_sample.py
```diff
@@ -0,0 +1,26 @@
...
+        }
+    )
+    testapp.patch_json(
+        curated_set_genome['@id'],
+        {
+            'samples': [tissue['@id']]
+        }
+    )
+    res = testapp.get(tissue['@id'])
+    assert res.json.get('file_sets') == [measurement_set['@id'], analysis_set_base['@id'], curated_set_genome['@id']]
```
gabdank, line 26: I am not sure you can assume the order in the list will be the same

gabdank (changes requested):

src/igvfd/tests/test_types_sample.py
```diff
@@ -0,0 +1,27 @@
...
+        }
+    )
+    testapp.patch_json(
+        curated_set_genome['@id'],
+        {
+            'samples': [tissue['@id']]
+        }
+    )
+    res = testapp.get(tissue['@id'])
+    assert set(res.json.get('file_sets')) == set(
```
gabdank, line 26: the right side of the equation should be `set(measurment_set['id'],,,,....`. there is no need for [] inside the set()

gabdank (changes requested):

src/igvfd/tests/test_types_sample.py
ian-whaling: this change throws and error:
>       assert set(res.json.get('file_sets')) == set(measurement_set['@id'], analysis_set_base['@id'], curated_set_genome['@id'])
E       TypeError: set expected at most 1 argument, got 3

set() is used for type casting, I don't think it defines a set, { } can define a set.

gabdank (approved):

PR #255, opened 2023-03-09
IGVF-469-multiome

ian-whaling:


src/igvfd/types/file_set.py
```diff
@@ -86,6 +86,34 @@ class MeasurementSet(FileSet):
...
+            },
+            'notSubmittable': True,
+        }
+    )
+    def related_multiome_datasets(self, request, accession, samples=None):
+        related_datasets = []
+        if samples:
+            for sample in samples:
+                sample_object = request.embed(sample, '@@object')
+                for file_set in sample_object['file_sets']:
```
gabdank, line 110: if sample_object has no 'file_sets' you will get an exception here. probably should use "get"

src/igvfd/types/file_set.py
```diff
@@ -86,6 +86,34 @@ class MeasurementSet(FileSet):
...
+                    if 'measurement-sets' in file_set and \
```
gabdank, line 111: are file_sets embedded in the sample?

gabdank (changes requested):

src/igvfd/types/file_set.py
ian-whaling: No, just a list of links. i.e. [/measurement-sets/IGVFDS0000AAAA/,/analysis-sets/IGVFDS1111BBBB/]

src/igvfd/types/file_set.py
ian-whaling: I will add this, but sample_object should always have 'file_sets' since it's reverse linked and this is checking all samples on a given measurement_set.

gabdank: it is status filtered and better to assume as little as possible

src/igvfd/types/file_set.py
```diff
@@ -86,6 +86,35 @@ class MeasurementSet(FileSet):
...
+        if samples:
+            related_datasets = []
+            for sample in samples:
+                sample_object = request.embed(sample, '@@object')
+                if sample_object.get('file_sets'):
+                    for file_set in sample_object.get('file_sets'):
+                        if 'measurement-sets' in file_set and \
+                            accession not in file_set and \
+                                file_set not in related_datasets:
+                            related_datasets.append(file_set)
```
gabdank, line 115: it took me a while to understand, but I think I do understand now. (1) I think for clarity "file_set" has to be renamed into "file_set_id". (2) I think that "replaced" objects loose their accession, so one cannot rely on the presence of accession, instead use @id, (3) probably good idea also to indicate the measurment-sets is not just part of the id , but specifically prefix.

gabdank (changes requested):

src/igvfd/types/file_set.py
```diff
@@ -86,6 +86,36 @@ class MeasurementSet(FileSet):
...
+            'items': {
+                'title': 'Related Multiome Dataset',
+                'description': 'Related dataset included in the multiome experiment this measurement set is a part of.',
+                'type': 'string',
+                'linkTo': 'MeasurementSet'
+            },
+            'notSubmittable': True,
+        }
+    )
+    def related_multiome_datasets(self, request, samples=None):
```
gabdank, line 105: I think this calculation should be applied ONLY if the measurment_set if multiome, i don't see this condition

gabdank (changes requested):

src/igvfd/types/file_set.py
ian-whaling: condition='multiome_size',
means the calculated property will only calculate the property if the measurement set has `multiome_size,` I also have a test that affirms this in test_types_measurement_set.py on line 18

gabdank (dismissed):

PR #256, opened 2023-03-09
IGVF-468-sample-sorted-fraction

KAndreeva:
Added sorted fraction to sample schema.

### DIFF
--- a/src/igvfd/schemas/changelogs/sample.md
+++ b/src/igvfd/schemas/changelogs/sample.md
@@ -1,5 +1,9 @@
 ## Changelog for *`sample.json`*
 
+### Release v7
+
+* Add `sorted_fraction`.
+
 ### Release v6
 
 * Add `file_sets`.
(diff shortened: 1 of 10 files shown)


src/igvfd/schemas/changelogs/sample.md
```diff
@@ -1,5 +1,9 @@
 ## Changelog for *`sample.json`*
 
+### Release v7
```
gabdank, line 3: I think we need to add "minor change" on the subclasses that got this new property? otherwise it is unclear on tittue or in vitro system it will be unclear from change log

gabdank (changes requested):

src/igvfd/schemas/changelogs/sample.md
KAndreeva: Done.

gabdank (dismissed):

gabdank (approved):

PR #261, opened 2023-03-13
IGVF-435-reconcile-human-donor

dougli1sqrd:
Fixing a typo in the description of references in mixins schema.

Previously it said `A reference that provide smore information about the object` and now it says `A reference that provides more information`.

### DIFF
--- a/src/igvfd/schemas/mixins.json
+++ b/src/igvfd/schemas/mixins.json
@@ -301,7 +301,7 @@
             "uniqueItems": true,
             "items": {
                 "title": "Reference",
-                "description": "A reference that provide smore information about the object.",
+                "description": "A reference that provides more information about the object.",
                 "type": "string",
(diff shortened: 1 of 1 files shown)


forresttanaka (dismissed):
OK, I got completely confused. I didn’t realize you had an igvfd ticket under the same ticket number as an igvf-ui ticket. I think this really needed a new ticket. I don’t think we use the `description` property yet, though we might in encoded.

hitz:
Should this PR be closed @gabdank @forresttanaka

hitz:
ticket is already merged?

forresttanaka:
It’s not merged. The igvf-ui branch of IGVF-435 is merged, but the igvfd one isn’t. I told Eric in the future we need different tickets for the igvfd and igvf-ui portions of one fix, because this is really confusing.

PR #266, opened 2023-03-15
IGVF-502-modification

jenjou:


gabdank (dismissed):

gabdank (changes requested):
I think we need to add search/config for modification (facets/columns, and probably more boost values) it would be also useful to create a ticket for UI modification (search and object page) should be tightly related to columns/facets definitions

src/igvfd/schemas/changelogs/biosample.md
```diff
@@ -1,5 +1,8 @@
 ## Changelog for *`biosample.json`*
 
+### Release v9
+
+* Add `modification`.
 
 ### Minor changes since Release v7
```
jenjou, line 7: I don't think we need a "minor changes since..." in the changelog for abstract classes but I didn't want to modify that in this ticket.

gabdank: agree

gabdank (approved):

PR #271, opened 2023-03-20
IGVF-482-remove-index-migration-script

keenangraham:


ottojolanki:
Just making sure you want to leave the script in `src/igvfd/commands/maybe_migrate_from_old_to_new_opensearch_indices.py` even though it is not expected to be used ever?

PR #277, opened 2023-03-24
IGVF-501-multiome-audits

ian-whaling:


gabdank (dismissed):

src/igvfd/tests/test_audit_measurement_set.py
```diff
@@ -0,0 +1,120 @@
...
+        }
+    )
+    testapp.patch_json(
+        measurement_set['@id'],
+        {
+            'multiome_size': 3
+        }
+    )
+    res = testapp.get(measurement_set_multiome['@id'] + '@@index-data')
+    assert res.json['audit'].get('WARNING', []) == []
```
gabdank, line 120: I don't think this test will work in general - as you can imagine other WARNINGs and the assertion here requires no WARNINGs

gabdank (changes requested):
I think assertion should be refined to specific error category, rather than to audit level

gabdank (approved):

PR #292, opened 2023-04-04
IGVF-535-add-s3-uri-to-files

ottojolanki:


cdk/infrastructure/config.py
```diff
@@ -57,7 +57,7 @@
                         'construct_id': 'Postgres',
                         'on': True,
                         'props': {
```
gabdank, line 59: @ottojolanki why do we start from a different db snapshot?

src/igvfd/types/file.py
```diff
@@ -81,6 +81,12 @@ def show_upload_credentials(request=None, context=None, upload_status=None):
     return request.has_permission('edit', context)
 
 
+def show_cloud_metadata(status=None, md5sum=None, file_size=None, restricted=None, no_file_available=None):
+    if restricted or not md5sum or not file_size or no_file_available:
```
gabdank, line 85: i don't think we have "restricted" in the file properties for IGVF

gabdank: same for "no_file_available"

gabdank (changes requested):

src/igvfd/tests/test_types_file.py
```diff
@@ -45,3 +46,27 @@ def test_types_file_set_external_sheet(sequence_data, root):
...
+        'maximum_read_length': 101,
+        'mean_read_length': 100,
+        'read_count': 23040138,
+        'file_size': 5495803,
+        'content_type': 'reads',
+        'sequencing_run': 1,
+        's3_uri': 's3://foo/bar/baz.fastq.gz'
+    }
+    with pytest.raises(webtest.app.AppError):
+        testapp.post_json('/sequence_data/', item, status=201).json['@graph'][0]
```
keenangraham, line 72: Seems a little strange to pull in the low-level `webtest` exception here. You won't really know what you're catching with `AppError`, could be anything. 

Also the `status=201` in the `post_json` call implies that you are expecting a successful response. Also not sure why `.json['@graph'][0]` is important here?

I think what you want to do is get rid of the `with pytest.raises(webtest.app.AppError):` block and simply post the item, expecting `status=422` (validation failure).

src/igvfd/tests/test_types_file.py
```diff
@@ -1,4 +1,5 @@
 import pytest
+import webtest
```
keenangraham, line 2: I don't think you need this, but, if you do, should import in the test's local scope.

src/igvfd/types/file.py
```diff
@@ -112,6 +112,23 @@ def href(self, request, file_format, accession):
             filename
         )
 
+    @calculated_property(
+        schema={
+            'title': 'S3 URI',
+            'description': 'The S3 URI of public file object.',
```
keenangraham, line 118: These files aren't public. I'm not sure in general if it's a great idea to expose our internal bucket names, when they can't be accessed publicly.

It's important to remember that we only added s3_uri field to ENCODE after we had switched to the public dataset. Somehow checkfiles worked without it.

On the plus side at least this is still a form of relying on the portal to tell checkfiles where to find a file (e.g. will be simpler if files in another S3 bucket or even a GC bucket/ANViL project ever need to be checked).

Unfortunate though that the streaming checkfiles didn't work by relying directly on the portal access abstraction, instead of having information about the underlying buckets. Now code and credentials will have to be duplicated/maintained across portal/checkfiles if there are ever additional buckets/cloud providers added.

keenangraham (changes requested):

src/igvfd/types/file.py
ottojolanki: Yea. Worrying about other cloud providers etc should be worried when they get added though.

gabdank (approved):

gabdank (approved):

src/igvfd/types/file.py
keenangraham: You should update the description though.

PR #298, opened 2023-04-11
IGVF-538-refseq-id-fix

ian-whaling:


jenjou (changes requested):
Need to change the "schema_version" to 2, otherwise looks good!

jenjou (approved):

PR #301, opened 2023-04-13
IGVF-474-remove-and-audit-plasmid-map-on-ConstructLibrary

iyoungworth:


src/igvfd/upgrade/file_set.py
```diff
@@ -33,3 +33,15 @@ def measurement_set_3_4(value, system):
     if 'protocol' in value:
         if value['protocol'] == 'https://www.protocols.io/':
             del value['protocol']
+
+
+@upgrade_step('construct_library', '1', '2')
+def construct_library_1_2(value, system):
+    # https://igvf.atlassian.net/browse/IGVF-474
+    plasmid_map = value.get('plasmid_map', '')
```
gabdank, line 41: I think it would be more appropriate to do `value.get('plasmid_map', None)`

src/igvfd/schemas/changelogs/construct_library.md
```diff
@@ -1,5 +1,9 @@
 ## Changelog for *`construct_library.json`*
 
+### Schema version 2
+
+* Remove `plasmid_map` property. A plasmid map should be submitted to  `documents` property instead.
```
gabdank, line 5: in changelogs, we are trying to be very "terse". I think we can drop the "property" from the first sentence. So something like: `Remove "plasmid_map". Use "documents.document_type" instead.`

gabdank (changes requested):

src/igvfd/upgrade/file_set.py
iyoungworth: Got it, fixed

src/igvfd/schemas/changelogs/construct_library.md
iyoungworth: I made the recommended change, saying "Use 'documents' instead." (I left out document_type to indicate that the whole plasmid map document must be attached)

gabdank (approved):

PR #317, opened 2023-04-26
IGVF-562-audit-inherit-multiome

ian-whaling:


gabdank (approved):

src/igvfd/tests/test_audit_measurement_set.py
```diff
@@ -130,6 +130,35 @@ def test_audit_related_multiome_datasets(
         error['category'] != 'inconsistent multiome metadata'
         for error in res.json['audit'].get('WARNING', [])
     )
```
gabdank, line 132: Ian, it looks like the test for multiome audit now is more than 100 lines long,,, i think we need to nreak it into multiple tests for readability

gabdank (changes requested):

gabdank (approved):

PR #320, opened 2023-04-27
IGVF-572 Audit for biosample NIH institutional certification.

KAndreeva:
Added audit for biosample NIH institutional certification.

### DIFF
--- a/src/igvfd/audit/biosample.py
+++ b/src/igvfd/audit/biosample.py
@@ -12,9 +12,9 @@ from .formatter import (
 def audit_biosample_nih_institutional_certification(value, system):
     '''Biosample objects must specify an NIH Institutional Certification required for human data.'''
     if ('nih_institutional_certification' not in value) and (value.get('taxa') == 'Homo sapiens'):
-        value_id = system.get('path')
+        sample_id = value.get('@id')
         detail = (
-            f'Biosample {audit_link(path_to_text(value_id), value_id)} '
-            f'is missing NIH institutional certification required for human data.'
+            f'Biosample {audit_link(path_to_text(sample_id), sample_id)} '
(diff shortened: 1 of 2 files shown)


src/igvfd/audit/biosample.py
```diff
@@ -0,0 +1,20 @@
...
+    audit_link,
+    path_to_text,
+)
+
+
+@audit_checker('Biosample', frame='object')
+def audit_biosample_nih_institutional_certification(value, system):
+    '''Biosample objects must specify an NIH Institutional Certification required for human data.'''
+    if ('nih_institutional_certification' not in value) and (value.get('taxa') == 'Homo sapiens'):
+        value_id = system.get('path')
```
gabdank, line 15: why not use "sample_id"? it would be more clear this way ID of what you are looking at. And why use system.get('path'), instead of `value.get('@id')`

src/igvfd/audit/biosample.py
```diff
@@ -0,0 +1,20 @@
...
+        detail = (
+            f'Biosample {audit_link(path_to_text(value_id), value_id)} '
+            f'is missing NIH institutional certification required for human data.'
```
gabdank, line 18: maybe "is missing NIH institutional certificate that is required for human samples"?

gabdank (changes requested):

src/igvfd/audit/biosample.py
KAndreeva: Done

src/igvfd/audit/biosample.py
KAndreeva: Done

gabdank (approved):

PR #321, opened 2023-04-27
IGVF-616-add-CRISPR-enum

iyoungworth:


gabdank (changes requested):
can you please add inserts for the new properties enums

gabdank (approved):

PR #328, opened 2023-05-02
IGVF-575-part-and-pooled-not-submittable-whole-organism

iyoungworth:


src/igvfd/schemas/whole_organism.json
```diff
@@ -68,7 +68,13 @@
     },
     "properties": {
         "schema_version": {
-            "default": "10"
+            "default": "11"
+        },
+        "part_of": {
+            "type": "null"
```
ian-whaling, line 74: Could you leave a comment explaining that this property should not be submitted.

ian-whaling (changes requested):

src/igvfd/schemas/whole_organism.json
ian-whaling: these properties*

iyoungworth: Good idea, added

ian-whaling (approved):

PR #332, opened 2023-05-04
IGVF-568-signal-file

ian-whaling:


src/igvfd/schemas/signal_file.json
```diff
@@ -0,0 +1,106 @@
...
+            "type": "boolean"
+        },
+        "normalized": {
+            "title": "Normalized",
+            "description": "Indicates if the signal file is normalized.",
+            "type": "boolean"
+        },
+        "start_view_position": {
+            "title": "Start View Position",
+            "description": "The starting position within the chromosome for viewing the signal.",
```
jenjou, line 94: We should probably mention that this is supposed to be a 0-based coordinate. Maybe "The 0-based coordinate for the default starting position when viewing the signal in a genome browser"?

ian-whaling: added

gabdank (approved):

PR #333, opened 2023-05-04
IGVF-584-ntr-treatment-audit

corismall:


src/igvfd/audit/treatment.py
```diff
@@ -0,0 +1,19 @@
...
+
+@audit_checker('Treatment', frame='object')
+def audit_treatment_term_id_check(value, system):
+    '''If treatment_term_id starts with NTR, flag treatment object with warning of new term request.'''
+    if 'treatment_term_id' in value:
+        term_id = value['treatment_term_id']
+        if term_id.startswith('NTR'):
+            treatment_id = value['@id']
+            detail = f'Treatment {audit_link(treatment_id, treatment_id)} has term_id {audit_link(term_id,term_id)}.'
+            yield AuditFailure('treatment term has been newly requested', detail, level='WARNING')
```
khzlin, line 19: What do you think of making the audit detail a bit more informative? Something like:
"This treatment term has been newly requested. It will be replaced with an UNIPROT or CHEBI term after it is added to the appropriate ontology database."

khzlin (changes requested):
See comment for minor change

src/igvfd/audit/treatment.py
corismall: sounds good, i'll update it

khzlin (approved):
looks good to me!

PR #337, opened 2023-05-08
IGVF-627-require-differentiation-properties

ian-whaling:
… time_post_factors_introduction_units


gabdank (approved):

gabdank (approved):

gabdank (changes requested):
I think we need to add changelog, especially since you are upgrading the version. something along the lines : in vitro systems A, B, C are required to specify blah blah

gabdank (dismissed):

gabdank (approved):

PR #355, opened 2023-05-16
IGVF-621-ccf-audit

jenjou:


src/igvfd/audit/tissue.py
```diff
@@ -18,3 +18,15 @@ def audit_tissue_ccf_id(value, system):
             f'is missing common coordinate framework identifier (CCF ID) required for human data.'
         )
         yield AuditFailure('missing ccf_id', detail, level='NOT_COMPLIANT')
+
+
+@audit_checker('Tissue', frame='object')
+def audit_tissue_ccf_id_nonhuman_sample(value, system):
+    '''Tissue objects must not specify a common coordinate framework identifier (CCF ID) unless the sample is from a human donor.'''
+    if ('ccf_id' in value) and (value.get('taxa') != 'Homo sapiens'):
```
gabdank, line 26: if it is a mixed donors sample you might not have taxa. I think this should be (value.get('taxa') -> (value.get('taxa', '')

src/igvfd/audit/tissue.py
```diff
@@ -18,3 +18,15 @@ def audit_tissue_ccf_id(value, system):
...
+        value_id = system.get('path')
+        detail = (
+            f'Tissue {audit_link(path_to_text(value_id), value_id)} '
+            f'has common coordinate framework identifier (CCF ID) but is not a human sample.'
```
gabdank, line 30: maybe in detail you can say "associated with non human donor" instead of non human sample?

src/igvfd/audit/tissue.py
```diff
@@ -18,3 +18,15 @@ def audit_tissue_ccf_id(value, system):
...
+        )
+        yield AuditFailure('unexpected ccf_id', detail, level='NOT_COMPLIANT')
```
gabdank, line 32: I think the link of non human to CCF is an ERROR

gabdank (changes requested):

gabdank (approved):

PR #364, opened 2023-05-18
IGVF-591-audit-whole-org-taxa-human

iyoungworth:


gabdank (dismissed):

src/igvfd/audit/whole_organism.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+@audit_checker('WholeOrganism', frame='object')
+def audit_whole_organism_human_taxa(value, system):
+    '''Flag whole organisms associated with taxa Homo sapiens.'''
+    if 'donors' in value:
+        donor_ids = value.get('donors')
+        taxa_dict = {}
+        for d in donor_ids:
+            donor_object = system.get('request').embed(d + '@@object?skip_calculated=true')
```
gabdank, line 18: if we are not using the donor that was human or not in the audit detail - why bother with dictionary? I think you can get away with set of the taxas?, and probably need to do the traversal - ONLY if the "taxa" calculated property is not in the sample?

gabdank (changes requested):
i think the traverasal shoudl happen only if the "taxa" is not calculated and then you don't need dict, but can use set instead?

src/igvfd/audit/whole_organism.py
iyoungworth: Got it, code is adjusted to use 'taxa' first if present. If not, donors are traversed and a set() of taxas is checked

PR #365, opened 2023-05-18
IGVF-573-audit-age-sample

khzlin:


src/igvfd/audit/primary_cell.py
```diff
@@ -0,0 +1,20 @@
...
+    audit_checker,
+    AuditFailure,
+)
+from .formatter import (
+    audit_link,
+    path_to_text,
+)
+
+
+@audit_checker('PrimaryCell', frame='object')
```
gabdank, line 11: this is the same function repeated 3 times, we need to refactor (put it in one place) and call it in three different places.

gabdank (changes requested):
needs to reduce code duplication

src/igvfd/tests/test_audit_primary_cell.py
```diff
@@ -0,0 +1,12 @@
+import pytest
+
+
+def test_audit_primary_cell_age(
+    testapp,
+    primary_cell
+):
+    res = testapp.get(primary_cell['@id'] + '@@index-data')
```
keenangraham, line 8: I think should call `@@audit` here instead of `@@index-data`?

khzlin: we've always used `@@index-data`, even in encoded. Should we be switching all to `@@audit`?

keenangraham: Yes, I don't see a reason to use `@@index-data` here.

src/igvfd/audit/primary_cell.py
khzlin: moving audit to biosample abstract level

khzlin: moving back to selective biosamples

gabdank (approved):

gabdank (approved):

PR #371, opened 2023-05-25
IGVF-650-add-ids-fuzzysearch

meenakshikagda:


cricketsloan (approved):
You have done what was asked.  I have opinions that I put in the ticket about which types maybe are not top hits, but this is opinion and I engaged Jennifer in that convo.

cricketsloan (dismissed):

cricketsloan (dismissed):
Did you confirm that accession still works for these properties.  I know that that is the plan.

cricketsloan (dismissed):

gabdank (approved):

PR #378, opened 2023-05-26
IGVF-683 Update MeasurementSet file table columns

forresttanaka:
Add some temporary test data.* I added a feature to SortTable so that you can conditionally hide a column. We had this with the encoded HTML table version of SortTable, but I hadn’t put that feature into this CSS-grid based one until now.

If we can’t get this into next week’s build, no big deal — it can go into the following week’s build.


mingjiecn (approved):

forresttanaka:
I was _supposed_ to PR the igvf-ui branch, not this one.

PR #391, opened 2023-06-07
IGVF-725-add-search-faq

keenangraham:


forresttanaka (dismissed):
I think the FAQ format is good. The one content change I can think of, though I’ll have Idan decide if we need that or not, is the difference between `TOP_HITS_ITEM_TYPES` and `DEFAULT_ITEM_TYPES` so that someone maintaining this in the future knows when to make the the same or different.

src/igvfd/searches/README.md
```diff
@@ -0,0 +1,43 @@
...
+**Q: Can I still search for the exact value if it's not included in `exact_searchable_fields`?**
+
+A: Yes, you can always search for exact values if you know the field that you want to use, even on fields that aren't included in either `fuzzy_searchable_fields` or `exact_searchable_fields`:
+
+* `?uuid=1e75d989-a438-4d77-a451-8a297fd3636e`
+* `?status=released`
+
+**Q: Why are certain terms, like `UUID`, searchable by default?**
+
+A: Certain high-level document values, such as UUIDs and unique keys, are copied by default into the `_exact` field in Opensearch. You can look at the mapping for a type to see if a specific field is copied into `_exact` or `_fuzzy` fields.
```
gabdank, line 27: can we add a path to the mappings example in the repo? i don't think we can assume everyone understands what "mapping" is

src/igvfd/searches/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Search configs:
+
+**Q: What are search configs?**
```
gabdank, line 31: can we add a path to the search configs in the repo? i don't think we can assume everyone understands what search configs are.

gabdank: similar to the link [`DEFAULT_ITEM_TYPES`](https://github.com/IGVF-DACC/igvfd/blob/b3714678fb6695a97a544eecb0afe403cdf149c9/src/igvfd/searches/defaults.py#L22-L60)

src/igvfd/searches/README.md
gabdank: similar to the link [`DEFAULT_ITEM_TYPES`](https://github.com/IGVF-DACC/igvfd/blob/b3714678fb6695a97a544eecb0afe403cdf149c9/src/igvfd/searches/defaults.py#L22-L60)

gabdank (changes requested):
can we add links?

gabdank (approved):

PR #393, opened 2023-06-07
IGVF-729-add-cas-species-to-modification

corismall:


src/igvfd/upgrade/modification.py
```diff
@@ -0,0 +1,8 @@
+from snovault import upgrade_step
+
+
+@upgrade_step('modification', '1', '2')
```
gabdank, line 4: add notes indicating that the upgrade to include Sp was performed automatically and was not part of original submission

gabdank (changes requested):
add notes in the upgrade

src/igvfd/upgrade/modification.py
```diff
@@ -0,0 +1,10 @@
...
+def modification_1_2(value, system):
+    # https://igvf.atlassian.net/browse/IGVF-729
+    if 'cas_species' not in value:
+        value['cas_species'] = 'Streptococcus pyogenes (Sp)'
+        value[
+            'notes'] = 'For upgrade, cas_species has been automatically designated as Streptococcus pyogenes (Sp), follow up with associated lab to check if upgrade is valid.'
```
gabdank, line 10: what about the possibility of previous notes content?

gabdank (changes requested):

src/igvfd/schemas/modification.json
```diff
@@ -92,6 +95,18 @@
             "description": "The product or catalog identifier provided following deposition to addgene.org.",
             "type": "string",
             "pattern": "^addgene:\\d{5,6}$"
+        },
+        "cas_species": {
+            "title": "Cas species",
+            "description": "The originating species of the Cas nuclease.",
+            "type": "string",
+            "default": "Streptococcus pyogenes (Sp)",
```
gabdank, line 103: if it is required, not sure we should default to anything

src/igvfd/tests/test_schema_modification.py
```diff
@@ -0,0 +1,36 @@
...
+                             {'cas_species': 'Streptococcus pyogenes (Sp)'})
+    assert res.status_code == 200
+    res = testapp.patch_json(modification['@id'],
+                             {'cas_species': ''}, expect_errors=True)
+    assert res.status_code == 422
+
+
+def test_read_cas_species(modification, testapp):
+    res = testapp.get(modification['@id'])
+    assert res.json['cas_species'] == 'Streptococcus pyogenes (Sp)'''
```
gabdank, line 15: i am not sure what is this testing? and it also ends with '''?

gabdank (changes requested):

src/igvfd/tests/test_schema_modification.py
corismall: I wanted to test the default assignment, but i realize that the fixture has `cas_species` explicitly assigned, and the test doesn't pass unless it is explicitly assigned.  I removed the ''' at the end.

src/igvfd/schemas/modification.json
corismall: makes sense, i can remove it

src/igvfd/tests/test_schema_modification.py
corismall: will remove now that i've removed default

gabdank (approved):

PR #406, opened 2023-06-15
IGVF-357-update-schema-making-docs

cricketsloan:
IGVF-357 updating the schema making documentation

### DIFF
--- a/docs/updating_schemas.md
+++ b/docs/updating_schemas.md
@@ -7,17 +7,20 @@ Guide to where to edit Source Code
 ----------------
 
 * **src/igvfd** Directory - contains all the Python and Javascript code for front and backends
-    * **audit** - Contains Python scripts that check JSON objects' metadata stored in the schema
-    * **schemas** - JSON schemas ([JSONSchema], [JSON-LD]) describing allowed types and values for all metadata objects
-    * **schemas/changelogs** - Schema change logs for documenting different versions of schemas
-    * **schemas/mixins.json** - Common schema property definitions for use with multiple schemas
-    * **tests** - Unit and integration tests
-    * **tests/data/inserts** - The sample data that comes up in a basic local instance.
-    * **tests/fixtures/schemas** - The sample Python objects to use with unit tests.  These Python function names are used as parameters in unit tests.
-    * **types** -  Business logic for dispatching URLs and producing the correct JSON
-    * **upgrade** - Python instructions for upgrading old objects to match the new schema
-    * **loadxl.py** - Python script that defines the schema objects to load
-    * **schema_format.py** - The format checker for accessions
+    * **/audits** - Contains Python scripts that are run post-submission to check JSON objects' metadata stored in the schema
+    * **/schemas** - JSON schemas ([JSONSchema], [JSON-LD]) describing allowed types and values for all metadata objects
+      * **/schemas/changelogs** - Schema change logs for documenting the changes made in each version of a schema
+      * **/schemas/mixins.json** - Common schema property definitions for use with multiple schemas
+    * **/tests** - Unit and integration tests
+      * **/tests/data/inserts** - The sample data that comes up in a basic local instance.
+      * **/tests/fixtures/schemas** - The sample Python objects to use with unit tests.  These Python function names are used as parameters in unit tests.
+    * **/types** -  Logic for dispatching URLs and producing the correct JSON.  This is where computed fields are specified.
+    * **/upgrade** - Python instructions for upgrading old objects to match the new schema
+    * **/loadxl.py** - Python script that defines the schema objects to load
+    * **/schema_format.py** - The format checker for accessions
+    * **/searches** - Directory that contains the configurations for search results
+      * **/searches/defaults.py** - This file contains a list that determines if a new schema is searched by default
+      * **/searches/configs** - Directory that has a file for each type that contains the list of fields returned in a search
 
 
 -----
@@ -25,7 +28,7 @@ Guide to where to edit Source Code
 Adding a new schema
 ----------------
 
-1. To add a new schema navigate to the **schemas** directory and make a new JSON file named after the object. Populate the file with basic schema definition:
+1. Add a new JSON file to the **schemas** directory named after the object (i.e. antibody.json). Populate the file with basic schema definition:
 
 
             {
@@ -34,6 +37,7 @@ Adding a new schema
                 "id": "/profiles/{metadata object}.json",
                 "$schema": "http://json-schema.org/draft-04/schema#",
                 "identifyingProperties": ["uuid"],
+                "required": [],
                 "additionalProperties": false,
                 "mixinProperties": [
                     { "$ref": "mixins.json#/schema_version" },
@@ -80,13 +84,23 @@ Adding a new schema
             }
 
 
-3. Identify all required properties to make an object and add to the "required" array. For example for treatment object type we might have the following properties:
+3. Identify all required properties for an object type and add them to the "required" array. For example for treatment object type we might have the following properties:
 
 
             "required": ["treatment_term_name", "treatment_type"]
 
+4. Add the "exact_searchable_fields" and "fuzzy_searchable_fields" properties using this [guide](https://github.com/IGVF-DACC/igvfd/tree/dev/src/igvfd/searches).
 
-4. In the **types** directory add a collection class for the object to define the rendering of the object.
+            "fuzzy_searchable_fields": [
+                 "name",
+                 "@type",
+                 "synonyms"
+            ],
+            "exact_searchable_fields": [
+                 "dbxrefs"
+            ], 
+
+5. In the **types** directory add a collection class for the object to define the rendering of the object.
 Refer to [object-lifecycle.md](https://github.com/IGVF-DACC/igvfd/blob/dev/docs/object_lifecycle.md) to understand object rendering. Example of basic collection definition for treatments:
 
 
@@ -102,7 +116,7 @@ Refer to [object-lifecycle.md](https://github.com/IGVF-DACC/igvfd/blob/dev/docs/
                 schema = load_schema('encoded:schemas/treatment.json')
 
 
-5. Within in a class add in  *embedding*, *reverse links*, and *calculated properties* as necessary.
+6. Within in a class add in  *embedding*, *reverse links*, and *calculated properties* as necessary.
 
     * *Embedding* - specifying the properties embeded in the object when specifying ```frame=object```, for construct we have:
 
@@ -125,7 +139,7 @@ Refer to [object-lifecycle.md](https://github.com/IGVF-DACC/igvfd/blob/dev/docs/
                 def title(self, term_name):
                     return term_name
 
-6. In ``loadxl.py`` add the new metadata object into the ```Order``` array, for example to add new object ```train.json```.
+7. In ``loadxl.py`` add the new metadata object into the ```Order``` array, for example to add new object ```train.json```.
 
             ORDER = [
                 'user',
@@ -137,7 +151,7 @@ Refer to [object-lifecycle.md](https://github.com/IGVF-DACC/igvfd/blob/dev/docs/
                 'train',
             ]
 
-7.  To load test fixtures of the new metadata object add them to ``/tests/conftest.py`` into the ```pytest_plugins``` array, for example to add new object ```train.json```.
+8.  To load test fixtures of the new metadata object add them to ``/tests/conftest.py`` into the ```pytest_plugins``` array, for example to add new object ```train.json```.
 
             pytest_plugins = [
                 'igvfd.tests.fixtures.database',
@@ -152,7 +166,7 @@ Refer to [object-lifecycle.md](https://github.com/IGVF-DACC/igvfd/blob/dev/docs/
                 'igvfd.tests.fixtures.schemas.train',
             ]
 
-8. Add in sample data to test the new schema in **tests** directory. Create a new JSON file in the **data/inserts** directory named after the new metadata object.
+9. Add in sample data to test the new schema in **tests** directory. Create a new JSON file in the **data/inserts** directory named after the new metadata object.
 This new object is an array of example objects that can successfully POST against the schema defined, for example:
 
             [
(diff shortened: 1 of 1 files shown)


cricketsloan:
I think that you can review the content of the mark down while I sort out why I am getting trailing white spaces.  This is mark down so it doesn't really matter

docs/updating_schemas.md
```diff
@@ -7,25 +7,28 @@ Guide to where to edit Source Code
 ----------------
 
 * **src/igvfd** Directory - contains all the Python and Javascript code for front and backends
```
keenangraham, line 9: No Javascript/frontend anymore.

docs/updating_schemas.md
```diff
@@ -7,25 +7,28 @@ Guide to where to edit Source Code
...
-    * **loadxl.py** - Python script that defines the schema objects to load
-    * **schema_format.py** - The format checker for accessions
+    * **/audits** - Contains Python scripts that are run post-submission to check JSON objects' metadata stored in the schema
+    * **/schemas** - JSON schemas ([JSONSchema], [JSON-LD]) describing allowed types and values for all metadata objects
+      * **/schemas/changelogs** - Schema change logs for documenting the changes made in each version of a schema
+      * **/schemas/mixins.json** - Common schema property definitions for use with multiple schemas
+    * **/tests** - Unit and integration tests
+      * **/tests/data/inserts** - The sample data that comes up in a basic local instance.
+      * **/tests/fixtures/schemas** - The sample Python objects to use with unit tests.  These Python function names are used as parameters in unit tests.
+    * **/types** -  Logic for dispatching URLs and producing the correct JSON.  This is where computed fields are specified.
```
keenangraham, line 17: Calculated properties?

docs/updating_schemas.md
```diff
@@ -102,7 +117,7 @@ Refer to [object-lifecycle.md](https://github.com/IGVF-DACC/igvfd/blob/dev/docs/
                 schema = load_schema('encoded:schemas/treatment.json')
 
 
-5. Within in a class add in  *embedding*, *reverse links*, and *calculated properties* as necessary.
+6. Within in a class add in  *embedding*, *reverse links*, and *calculated properties* as necessary.
```
keenangraham, line 120: `Within in` -> `Within`

docs/updating_schemas.md
```diff
@@ -102,7 +117,7 @@ Refer to [object-lifecycle.md](https://github.com/IGVF-DACC/igvfd/blob/dev/docs/
...
 
     * *Embedding* - specifying the properties embeded in the object when specifying ```frame=object```, for construct we have:
```
keenangraham, line 122: This doesn't sound right. `frame=object` doesn't resolve links to embedded objects, `frame=embedded` does.

docs/updating_schemas.md
keenangraham: There's also slim embedding using `embedded_with_frame`.

docs/updating_schemas.md
```diff
@@ -181,15 +195,49 @@ This new object is an array of example objects that can successfully POST agains
                     res = testapp.post_json('/user', item)
                     return testapp.get(res.location).json
```
keenangraham, line 196: Post and patch will usually GET the object after creating it and return it, so this GET is redundant.

keenangraham (dismissed):

keenangraham:
> I think that you can review the content of the mark down while I sort out why I am getting trailing white spaces. This is mark down so it doesn't really matter. Anyway, i think you will want changes on the text.

You should be able to lint locally by installing https://github.com/IGVF-DACC/igvfd#automatic-linting and running `pre-commit run --all-files` then committing whatever changes it makes.

docs/updating_schemas.md
cricketsloan: Good catch.  I wonder if we should do a more fine toothed review of the whole doc.  I was focused on adding the new steps.

docs/updating_schemas.md
cricketsloan: @keenangraham now that we have slim_embedding should we ever use embedding?

keenangraham: Still kind of think `embedding` should be the default since that was the original intention of the system, though maybe it doesn't matter. I don't fully understand the consequences of using slim embedding everywhere in the big context of our system (e.g. JSON-LD). I know the Opensearch mapping code still assumes all of the fields are embedded from another object, but maybe that's something that can just be changed. The original intention was to break circular dependencies and avoid expensive calculated properties (e.g. an experiment that wants some information from a reference dataset that embeds the experiment itself and 1000 other experiments). I'm not sure if that means it should be used everywhere.

keenangraham (approved):

docs/updating_schemas.md
```diff
@@ -34,6 +36,7 @@ Adding a new schema
                 "id": "/profiles/{metadata object}.json",
                 "$schema": "http://json-schema.org/draft-04/schema#",
```
jenjou, line 37: This should be updated to "https://json-schema.org/draft/2020-12/schema".

docs/updating_schemas.md
```diff
@@ -3,29 +3,31 @@ Making changes to schemas
...
------
+* **src/igvfd** Directory - contains all the Python code and JSON schemas for adding new types to the igvfd backend.
+    * **/audits** - Contains Python scripts that are run post-submission to check JSON objects' metadata stored in the schema
+    * **/schemas** - JSON schemas ([JSONSchema], [JSON-LD]) describing allowed types and values for all metadata objects
+      * **/schemas/changelogs** - Schema change logs for documenting the changes made in each version of a schema
+      * **/schemas/mixins.json** - Common schema property definitions for use with multiple schemas
+    * **/tests** - Unit and integration tests
+      * **/tests/data/inserts** - The sample data that comes up in a basic local instance.
+      * **/tests/fixtures/schemas** - The sample Python objects to use with unit tests.  These Python function names are used as parameters in unit tests.
+    * **/types** -  Logic for dispatching URLs and producing the correct JSON.  This is where calculated properties are specified.
```
jenjou, line 19: There's a double space in this line.

docs/updating_schemas.md
```diff
@@ -3,29 +3,31 @@ Making changes to schemas
...
+      * **/searches/configs** - Directory that has a file for each type that contains the list of fields returned in a search
 
 Adding a new schema
 ----------------
 
-1. To add a new schema navigate to the **schemas** directory and make a new JSON file named after the object. Populate the file with basic schema definition:
+1. Add a new JSON file to the **schemas** directory named after the object (i.e. antibody.json). Populate the file with basic schema definition:
 
 
             {
```
jenjou, line 33: Should we add an entry in this block for "dependentSchemas"?

jenjou:
In 2.Add appropriate properties in the "properties" example block, I think we should update the example property names to use title case, since that's the format we're following for IGVF.

docs/updating_schemas.md
cricketsloan: fixed

docs/updating_schemas.md
cricketsloan: Likely, do you have thoughts on what we should add?

cricketsloan:
@jenjou said "In 2.Add appropriate properties in the "properties" example block, I think we should update the example property names to use title case, since that's the format we're following for IGVF."

I am confused as it seems that we are still using snake case for the property names see example here: https://github.com/IGVF-DACC/igvfd/blob/IGVF-357-schema-making-docs/src/igvfd/schemas/award.json

jenjou:
> @jenjou said "In 2.Add appropriate properties in the "properties" example block, I think we should update the example property names to use title case, since that's the format we're following for IGVF."
> 
> I am confused as it seems that we are still using snake case for the property names see example here: https://github.com/IGVF-DACC/igvfd/blob/IGVF-357-schema-making-docs/src/igvfd/schemas/award.json

Sorry, my comment was a bit ambiguous!
I meant the "title"  in each property could be updated, changing "Example string" to "Example String" so both words are capitalized.

docs/updating_schemas.md
jenjou: Would it be OK to leave it blank, as we did with the "required" entry?
Otherwise, something like the below is probably the simplest form of dependency we have:
```
"duration": {
            "required": [
                "duration_units"
            ],
            "comment": "duration_units must be specified if duration is specified"
        },
```

docs/updating_schemas.md
cricketsloan: removed

docs/updating_schemas.md
cricketsloan: changed

docs/updating_schemas.md
cricketsloan: fixed

docs/updating_schemas.md
jenjou: Resolved by adding step 4 below

docs/updating_schemas.md
```diff
@@ -102,12 +129,16 @@ Refer to [object-lifecycle.md](https://github.com/IGVF-DACC/igvfd/blob/dev/docs/
                 schema = load_schema('encoded:schemas/treatment.json')
 
 
-5. Within in a class add in  *embedding*, *reverse links*, and *calculated properties* as necessary.
+7. Within a class add in  *embedding*, *slim_embedding*, *reverse links*, and *calculated properties* as necessary.
```
jenjou, line 132: Double space

docs/updating_schemas.md
```diff
@@ -102,12 +129,16 @@ Refer to [object-lifecycle.md](https://github.com/IGVF-DACC/igvfd/blob/dev/docs/
...
 
-    * *Embedding* - specifying the properties embeded in the object when specifying ```frame=object```, for construct we have:
+    * *Embedding* - specifying the properties embeded in the object when specifying ```frame=embedded```, for construct we have:
 
                 embedded = ['target']
 
+    * *Slim embedding* - specifying select properties from hierachically linked objects.  For example, if you would like to embed only the term_name of a biosample.
```
jenjou, line 138: Double space

docs/updating_schemas.md
```diff
@@ -167,7 +197,7 @@ This new object is an array of example objects that can successfully POST agains
                     "uuid": "0137a084-57af-4f69-b756-d6a920393fde"
                 }
```
jenjou, line 199: Can we add the closing bracket here ] ?

docs/updating_schemas.md
```diff
@@ -179,17 +209,52 @@ This new object is an array of example objects that can successfully POST agains
...
 
             accession_re = re.compile(r'^IGVF(FI|DS|SR|AB|SM|BS|DO|GM|LB|PL|AN)[0-9][0-9][0-9][A-Z][A-Z][A-Z]$')
             est_accession_re = re.compile(r'^TST(FI|DS|SR|AB|SM|BS|DO|GM|LB|PL|AN)[0-9][0-9][0-9]([0-9][0-9][0-9]|[A-Z][A-Z][A-Z])$')
 
-12.  Add a change log markdown file for the new schema to the **schemas/changelogs** directory.
+14.  Add a change log markdown file for the new schema to the **schemas/changelogs** directory.
+
+            ## Changelog for *`award.json`*
+
+15. To make sure that the objects of that type are searched in unspecified searches add the item to TOP_HITS_ITEM_TYPES in **igvfd/src/igvfd/searches/defaults.py**.  For example to add a type for train sets
```
jenjou, line 226: Double space

docs/updating_schemas.md
```diff
@@ -167,7 +197,7 @@ This new object is an array of example objects that can successfully POST agains
                     "uuid": "0137a084-57af-4f69-b756-d6a920393fde"
                 }
 
-9. Add in fixtures to test the new schema in **tests** directory. Create a new .py file in the **fixtures/schemas** directory named after the new metadata object. Fixtures may be used to validate expected schema behavoir with tests defined in test files in **tests** directory.
+11. Add in fixtures to test the new schema in **tests** directory. Create a new .py file in the **fixtures/schemas** directory named after the new metadata object. Fixtures may be used to validate expected schema behavoir with tests defined in test files in **tests** directory.
```
jenjou, line 200: Typo in behavior

docs/updating_schemas.md
```diff
@@ -72,21 +75,45 @@ Adding a new schema
...
+                    "title": "Example String with Pattern",
                     "description": "An example of a property that must match a pattern",
                     "type": "string",
                     "pattern": "^[\\S\\s\\d\\-\\(\\)\\+]+$"
                 }
             }
 
 
-3. Identify all required properties to make an object and add to the "required" array. For example for treatment object type we might have the following properties:
+3. Identify all required properties for an object type and add them to the "required" array. Identify all of the identifying properties and add them to the "identifyingProperties" array.  For example for treatment object type we might have the following properties:
```
jenjou, line 86: Double space

docs/updating_schemas.md
```diff
@@ -3,37 +3,40 @@ Making changes to schemas
...
------
+* **src/igvfd** Directory - contains all the Python code and JSON schemas for adding new types to the igvfd backend.
+    * **/audits** - Contains Python scripts that are run post-submission to check JSON objects' metadata stored in the schema
+    * **/schemas** - JSON schemas ([JSONSchema], [JSON-LD]) describing allowed types and values for all metadata objects
+      * **/schemas/changelogs** - Schema change logs for documenting the changes made in each version of a schema
+      * **/schemas/mixins.json** - Common schema property definitions for use with multiple schemas
+    * **/tests** - Unit and integration tests
+      * **/tests/data/inserts** - The sample data that comes up in a basic local instance.
+      * **/tests/fixtures/schemas** - The sample Python objects to use with unit tests.  These Python function names are used as parameters in unit tests.
+    * **/types** - Logic for dispatching URLs and producing the correct JSON.  This is where calculated properties are specified.
```
jenjou, line 19: Double space

cricketsloan: @jenjou Are the double spaces that you are finding the ones after the end of a sentence?  Or are are there mid-sentence ones?  Is the IGVF text format to drop the spacing after the period?

jenjou: I think I only saw them between (after the end) of a sentence. I'm used to having a single space -- I guess double space works, but we seem to have more single than double spaces so I thought it should be consistent.

forresttanaka: I see two spaces between sentences here. Markdown-display packages always strip spaces between sentences, so you can put in 1000 spaces after a period and Markdown strips it down to one. But I agree with Jennifer that keeping it consistent in the source as one space between sentences is nice. You’ll find decades of civil wars fought over whether one space or two between sentences was correct. I’m on the [one-space](https://www.cultofpedagogy.com/two-spaces-after-period/) battalion.

PR #408, opened 2023-06-15
IGVF-727-add-delivery-to-measurement

KAndreeva:
Added delivery property to the measurement set object.

### DIFF
--- a/src/igvfd/tests/test_schema_measurement_set.py
+++ b/src/igvfd/tests/test_schema_measurement_set.py
@@ -54,14 +54,14 @@ def test_seqspec_pattern(testapp, measurement_set):
     assert res.status_code == 422
 
 
-def test_(testapp, measurement_set):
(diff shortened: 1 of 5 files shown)


src/igvfd/tests/test_schema_measurement_set.py
```diff
@@ -52,3 +52,17 @@ def test_seqspec_pattern(testapp, measurement_set):
         expect_errors=True
     )
     assert res.status_code == 422
+
+
+def test_(testapp, measurement_set):
```
gabdank, line 57: the name of the function is incomplete

gabdank (changes requested):
test function name wrong

src/igvfd/tests/test_schema_measurement_set.py
KAndreeva: Updated

gabdank (approved):

PR #411, opened 2023-06-19
IGVF-736-non-vitual-samples-audit

KAndreeva:
Added an audit to flag non-virtual samples linked to virtual samples.

### DIFF
--- a/src/igvfd/audit/sample.py
+++ b/src/igvfd/audit/sample.py
@@ -55,3 +55,48 @@ def audit_sample_virtual_donor_check(value, system):
             detail = (f'The sample {audit_link(sample_id, sample_id)} is linked to virtual donor(s):'
                       f'{[audit_link(path_to_text(d_id),d_id) for d_id in donors_error]}')
             yield AuditFailure('inconsistent sample metadata', detail, level='ERROR')
+
+
+@audit_checker('Sample', frame='object')
+def audit_non_virtual_sample_linked_to_virtual_sample(value, system):
+    '''Non-virtual samples should not be linked to virtual samples'''
+    is_virtual = value.get('virtual', False)
+    # set up category for audit created later
+    if is_virtual:
+        audit_category = 'virtual sample linked to non-virtual sample'
+        audit_detail_body = 'is virtual'
+        audit_detail_end = 'that is not virtual'
+    else:
+        audit_category = 'non-virtual sample linked to virtual sample'
+        audit_detail_body = 'is not virtual'
+        audit_detail_end = 'that is virtual'
+    value_id = system.get('path')
+    current_type_name = value['@type'][0]
+    # list of properties to check:  Add to this list for additional linked objects with virtual property
+    properties_to_check = ['sorted_fraction', 'part_of', 'pooled_from', 'originated_from']
+    for current_property_name in properties_to_check:
+        current_property_value = value.get(current_property_name, None)
+        if current_property_value:
+            # check if property is a list
+            if isinstance(current_property_value, list):
+                # the property is a list of linked samples
+                for current_property_id in current_property_value:
+                    linked_data = system.get('request').embed(current_property_id + '@@object?skip_calculated=true')
+                    if linked_data.get('virtual', False) != is_virtual:
+                        detail = (
+                            f'{current_type_name} {audit_link(path_to_text(value_id), value_id)} '
+                            f'{audit_detail_body} and has a linked {current_property_name} '
+                            f'({audit_link(path_to_text(current_property_id), current_property_id)}) {audit_detail_end}.'
+                        )
+                        yield AuditFailure(audit_category, detail, level='ERROR')
+            else:
+                # the property is a single linked sample
+                current_property_id = current_property_value
+                linked_data = system.get('request').embed(current_property_id + '@@object?skip_calculated=true')
+                if linked_data.get('virtual', False) != is_virtual:
+                    detail = (
+                        f'{current_type_name} {audit_link(path_to_text(value_id), value_id)} '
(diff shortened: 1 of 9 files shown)


src/igvfd/audit/sample.py
```diff
@@ -35,3 +35,53 @@ def audit_sample_sorted_fraction_parent_child_check(value, system):
...
+
+
+def get_virtual_sample_failures(
+    system,
+    sample_id,
+    sample_is_virtual,
+    linked_sample_id
+):
+    if not linked_sample_id:
+        return None
```
gabdank, line 69: this will never happen, you can remove

src/igvfd/tests/fixtures/schemas/primary_cell.py
```diff
@@ -8,19 +8,21 @@ def primary_cell(testapp, other_lab, award, human_donor, sample_term_pluripotent
...
-        'biosample_term': sample_term_pluripotent_stem_cell['@id']
+        'biosample_term': sample_term_pluripotent_stem_cell['@id'],
+        'virtual': True
     }
     return testapp.post_json('/primary_cell', item, status=201).json['@graph'][0]
 
 
 @pytest.fixture
-def pooled_from_primary_cell(testapp, lab, award, source, human_donor, sample_term_pluripotent_stem_cell):
+def pooled_from_primary_cell(testapp, lab, award, source, human_donor, sample_term_pluripotent_stem_cell, primary_cell):
```
gabdank, line 18: why do we add "primary_cell" to the parameters?

src/igvfd/tests/fixtures/schemas/primary_cell.py
```diff
@@ -8,19 +8,21 @@ def primary_cell(testapp, other_lab, award, human_donor, sample_term_pluripotent
         'lab': other_lab['@id'],
         'source': other_lab['@id'],
         'donors': [human_donor['@id']],
-        'biosample_term': sample_term_pluripotent_stem_cell['@id']
+        'biosample_term': sample_term_pluripotent_stem_cell['@id'],
+        'virtual': True
```
gabdank, line 12: why do you make primary_cell virtual?

gabdank (changes requested):

src/igvfd/audit/sample.py
```diff
@@ -35,3 +35,46 @@ def audit_sample_sorted_fraction_parent_child_check(value, system):
...
+                    if audit_failure:
+                        yield audit_failure
+            else:
+                audit_failure = raise_audit_failure_if_not_match_virtual(
+                    system, sample_id, sample_is_virtual, current_property_value, current_property_name)
+                if audit_failure:
+                    yield audit_failure
+
+
+def raise_audit_failure_if_not_match_virtual(system, sample_id, sample_is_virtual, current_property_id, current_property_name):
```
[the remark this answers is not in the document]
KAndreeva: Function name changed.

src/igvfd/audit/sample.py
KAndreeva: This change causes the unit test to fail.  The lines on 46 to 48 potentially return None in the list of linked IDs if the property is not present in the sample.

src/igvfd/tests/fixtures/schemas/primary_cell.py
KAndreeva: This is needed to support my unit test named: test_non_virtual_sample_linked_to_virtual_sample_with_single_property

The fixture primary_cell_with_part_of_virtual_true has part_of linked to the fixture primary_cell.  primary_cell_with_part_of_virtual_true is set to virtual False.  primary_cell is set to virtual True.

src/igvfd/tests/fixtures/schemas/primary_cell.py
KAndreeva: Removed.

src/igvfd/tests/fixtures/schemas/primary_cell.py
gabdank: this fixture is being used by other tests, are you sure they are ok with it becoming virtual?

src/igvfd/audit/sample.py
gabdank: if the linked IDs are not present then links_to_check should be empty list?

src/igvfd/tests/fixtures/schemas/primary_cell.py
gabdank: this fixture is being used by other tests, are you sure they are ok with it becoming virtual?

gabdank (changes requested):

src/igvfd/tests/fixtures/schemas/primary_cell.py
KAndreeva: All unit tests passed with this change, but I can make another fixture if you want?

src/igvfd/audit/sample.py
KAndreeva: But we are populating them with value None if the property is not present. Each entry will be an element of the list.

src/igvfd/tests/fixtures/schemas/primary_cell.py
KAndreeva: I have removed the virtual property from the primary_cell fixture and patched it in the unit test instead.

src/igvfd/audit/sample.py
KAndreeva: I tried the unit tests with these lines removed, and the test fails.

gabdank (approved):

gabdank (approved):

PR #412, opened 2023-06-19
IGVF-719-remove-unused-files-objects

khzlin:


gabdank (changes requested):
if you look in types/file.py you will find ReferenceData and SequenceData - please remove from there, and form any other place in the repo

gabdank (dismissed):

PR #414, opened 2023-06-19
IGVF-735-audit-samples-virtual-donors

corismall:


src/igvfd/audit/sample.py
```diff
@@ -35,3 +35,33 @@ def audit_sample_sorted_fraction_parent_child_check(value, system):
         )
         if prop_errors != '':
             yield AuditFailure('inconsistent sorted fraction metadata', detail, level='ERROR')
+
+
+@audit_checker('Sample', frame='object')
+def audit_sample_virtual_donor_check(value, system):
+    '''Non-virtual samples should not be linked to virtual donors.'''
+    if 'virtual' and 'donors' in value:
```
gabdank, line 43: if the sample is virtual - we should skip the check - it is wasteful to traverse donors if we already know the sample is virtual.

src/igvfd/audit/sample.py
```diff
@@ -35,3 +35,33 @@ def audit_sample_sorted_fraction_parent_child_check(value, system):
...
+        sample_id = value['@id']
+        sample_virtual = value['virtual']
+        donor_ids = value.get('donors')
+        donors_error = []
+        for d in donor_ids:
+            donor_object = system.get('request').embed(d + '@@object')
```
gabdank, line 49: i think you can also skip calculated.

src/igvfd/audit/sample.py
```diff
@@ -35,3 +35,33 @@ def audit_sample_sorted_fraction_parent_child_check(value, system):
...
+            donor_id = donor_object.get('@id')
+            donor_virtual = donor_object.get('virtual', False)
+            if (sample_virtual == True) and (donor_virtual == True):
+                pass
+
+            if (sample_virtual == True) and (donor_virtual == False):
+                pass
+
```
gabdank, line 57: don't need these if you are not in the loop in the first place (See my top comment)

src/igvfd/tests/test_audit_sample.py
```diff
@@ -27,3 +27,33 @@ def test_audit_sample_sorted_fraction_parent_child_check(
...
+    testapp.patch_json(
+        tissue['@id'],
+        {
+            'virtual': False,
+            'donors': [human_donor['@id'], rodent_donor['@id']]
+        }
+    )
+    res = testapp.get(tissue['@id'] + '@@index-data')
+    assert any(
+        error['category'] == 'non-virtual sample linked to virtual donor'
```
gabdank, line 57: I think the category could be more general: "inconsistent sample metadata"

gabdank (changes requested):

src/igvfd/audit/sample.py
corismall: i had that initially, but i couldn't produce `detail` object; skipping calculated i couldn't get value['@id'].

src/igvfd/audit/sample.py
```diff
@@ -35,3 +35,23 @@ def audit_sample_sorted_fraction_parent_child_check(value, system):
         )
         if prop_errors != '':
             yield AuditFailure('inconsistent sorted fraction metadata', detail, level='ERROR')
+
+
+@audit_checker('Sample', frame='object')
+def audit_sample_virtual_donor_check(value, system):
+    '''Non-virtual samples should not be linked to virtual donors.'''
+    if ('donors' in value) and (value['virtual'] == False):
```
gabdank, line 43: better value.get('virtual', False)

gabdank (changes requested):
minor suggestion

gabdank (approved):

PR #422, opened 2023-06-22
IGVF-724-create-Prediction-File-Set

iyoungworth:


src/igvfd/schemas/prediction.json
```diff
@@ -0,0 +1,70 @@
+{
+    "title": "Prediction",
+    "description": "Schema for submitting a file set of prediction files generated by CAMP.",
```
gabdank, line 3: Not sure we want to use abbreviations like CAMP. 

Maybe just "set of computational predictions"?

gabdank (changes requested):

src/igvfd/searches/configs/prediction.py
```diff
@@ -0,0 +1,47 @@
+from snovault.elasticsearch.searches.configs import search_config
+
+
+@search_config(
+    name='Prediction'
+)
+def cell_line():
```
gabdank, line 7: not a cell line

gabdank (changes requested):
not a cell line in search config

src/igvfd/searches/configs/prediction.py
iyoungworth: Same copy paste error as in auxiliary set, sorry about this. Fixed.

gabdank (dismissed):

gabdank (approved):

PR #424, opened 2023-06-27
IGVF-717-analysis-step

jenjou:


src/igvfd/schemas/analysis_step.json
```diff
@@ -0,0 +1,138 @@
...
+            "type": "string",
+            "linkTo": "Workflow"
+        },
+        "parents": {
+            "title": "Parents",
+            "description": "The precursor steps.",
+            "uniqueItems": true,
+            "type": "array",
+            "items": {
+                "description": "A precusor step.",
```
iyoungworth, line 75: Minor typo

src/igvfd/schemas/analysis_step.json
```diff
@@ -0,0 +1,138 @@
...
+                "enum": [
+                    "alignment",
+                    "file format conversion",
+                    "signal generation"
+                ]
+            }
+        },
+        "step_label": {
+            "title": "Step Label",
+            "description": "Unique lowercased label of the analysis step using a naming system where the pipeline, software used, the purpose is described, e.g. rampage-grit-peak-calling-step",
```
iyoungworth, line 52: Should 'Pipeline' be 'Workflow'? Or maybe assay, as per the example?
I want to make sure if I should read this as a list of "possible options" for how a step might be named, or a suggestion that 3 things should be given in this order

src/igvfd/schemas/analysis_step.json
```diff
@@ -0,0 +1,138 @@
...
+        "title",
+        "step_label",
+        "analysis_step_types",
+        "input_content_types",
+        "workflow"
+    ],
+    "identifyingProperties": [
+        "uuid",
+        "aliases"
+    ],
```
iyoungworth, line 17: I think the calculated workflow-step-name that gets calculated will need to be added to this list in order to be used as an identifier

iyoungworth (changes requested):
You'll need to add a searches/defaults.py file and list this schema in searches/configs

src/igvfd/schemas/analysis_step.json
jenjou: My interpretation was that it should include 3 items in that order. That being said, do we still need the pipeline's title if we are going to calculate it to generate the "name"?

iyoungworth: Hm. Maybe not required for greater uniqueness, but I think including the assay or similar "overall" term is nice for human readable context. It does make it longer, though. I can see a case either way, I leave it to your judgment

src/igvfd/schemas/analysis_step.json
```diff
@@ -0,0 +1,139 @@
...
+            "$ref": "mixins.json#/basic_item"
+        },
+        {
+            "$ref": "mixins.json#/attribution"
+        },
+        {
+            "$ref": "mixins.json#/standard_status"
+        }
+    ],
+    "dependentSchemas": {},
```
gabdank, line 31: do we need empty dependentSchemas list?

src/igvfd/schemas/analysis_step.json
```diff
@@ -0,0 +1,139 @@
...
+            "type": "array",
+            "uniqueItems": true,
+            "items": {
+                "title": "Output Content Type",
+                "description": "A content type produced as output by the analysis step.",
+                "$merge": "#/$defs/content_types"
+            }
+        }
+    },
+    "$defs": {
```
gabdank, line 105: do we want to do something more extensive - like collecting content types from all the file subtypes?

gabdank (changes requested):

src/igvfd/schemas/analysis_step.json
jenjou: I wasn't sure how to do that. Is there a way I could reference content types in other schemas?

src/igvfd/schemas/analysis_step.json
```diff
@@ -0,0 +1,130 @@
...
+    "title": "Analysis Step",
+    "description": "Schema for submitting a step in a computational analysis workflow.",
+    "$id": "/profiles/analysis_step.json",
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "type": "object",
+    "required": [
+        "title",
+        "step_label",
+        "analysis_step_types",
+        "input_content_types",
```
gabdank, line 11: should not output_types be required as well?

gabdank (changes requested):
not sure why the outputs are not required like the inputs

src/igvfd/schemas/analysis_step.json
jenjou: That makes sense, will update

iyoungworth (approved):
This looks good. I think QA should include checking that 'name' works as an identifying property, that the expected content_types all show up, and the regexes in the label and title properties

PR #429, opened 2023-07-01
IGVF-794-rename-source-in-reference-file

corismall:


src/igvfd/tests/test_upgrade_reference_file.py
```diff
@@ -0,0 +1,8 @@
+import pytest
+
+
+def test_reference_file_upgrade_2_3(upgrader, reference_file_v2):
+    value = upgrader.upgrade('reference_file', reference_file_v2, current_version='2', target_version='3')
+    assert 'source' not in value
```
gabdank, line 6: i think you need to validate content of source_url vs content of source before upgrade

gabdank (changes requested):

src/igvfd/tests/test_upgrade_reference_file.py
corismall: like this? `assert reference_file_v2['source'] == value['source_url']`

gabdank (approved):

PR #430, opened 2023-07-06
IGVF-847-add-audit-doc-endpoint

ottojolanki:


docker/pyramid/Dockerfile
```diff
@@ -56,12 +60,22 @@ COPY --chown=igvfd:igvfd src/igvfd/schema_formats.py src/igvfd/schema_formats.py
 
 COPY --chown=igvfd:igvfd src/igvfd/ontology.py src/igvfd/ontology.py
 
+COPY --chown=igvfd:igvfd src/igvfd/audit/* src/igvfd/audit/
```
keenangraham, line 63: Can we put this as late a possible? Think audits change more than the stuff below. Maybe even after `RUN make-reference-db`?

src/igvfd/__init__.py
```diff
@@ -55,6 +57,15 @@ def favicon(request):
 
     config.add_view(favicon, route_name='favicon.ico')
 
+    def auditdoc(request):
```
keenangraham, line 60: I'm not sure this is necessary, and I think we want to avoid reading from disk on every request.

src/igvfd/__init__.py
```diff
@@ -47,6 +47,8 @@ def static_resources(config):
         favicon_path = '/%s%s' % (config.route_prefix, favicon_path)
     config.add_route('favicon.ico', 'favicon.ico')
 
+    config.add_route('auditdoc', '/auditdoc')
```
keenangraham, line 50: Can you create a /doc folder in the static folder in the package itself instead? https://github.com/IGVF-DACC/igvfd/tree/dev/src/igvfd/static Then https://{url}/static/doc/auditdoc should resolve it without having to create a route? For example right now this statically serves the favicon directly from folder I think: https://igvfd-dev.demo.igvf.org/static/img/favicon.ico

src/igvfd/commands/make_audit_docstring_json.py
```diff
@@ -0,0 +1,70 @@
...
+    audit_function_full_names = []
+    for audit_module in AUDIT_MODULES_TO_PROCESS:
+        module_name = audit_module.__name__
+        function_names_in_module = get_audit_function_names_from_module(audit_module)
+        for function_name in function_names_in_module:
+            audit_function_full_names.append(module_name + '.' + function_name)
+    audit_docstring_dict = {}
+    for function_name in audit_function_full_names:
+        audit_docstring_dict.update(get_docstring_dict_from_function_name(function_name))
+    with open('/static/json/auditdoc.json', 'w') as audit_json:
```
keenangraham, line 65: Can you just dump this into the `/igvfd/src/static/...` folder instead? Then the existing static route would just serve it I think: https://github.com/IGVF-DACC/igvfd/blob/cbd653c35060cbd9cc20cc273dea22b670314117/src/igvfd/__init__.py#L42

docker/pyramid/Dockerfile
```diff
@@ -34,6 +34,10 @@ RUN mkdir -p $REFERENCE_DATABASE_DIRECTORY
 
 RUN chown -R igvfd:igvfd $REFERENCE_DATABASE_DIRECTORY
 
+RUN mkdir -p /static/json
```
keenangraham, line 37: I think maybe we want to use the package static folder instead.

docker/pyramid/Dockerfile
```diff
@@ -56,12 +60,22 @@ COPY --chown=igvfd:igvfd src/igvfd/schema_formats.py src/igvfd/schema_formats.py
...
 
+COPY --chown=igvfd:igvfd src/igvfd/commands/make_audit_docstring_json.py src/igvfd/commands/make_audit_docstring_json.py
+
 RUN make-reference-db
 
+RUN touch /static/json/auditdoc.json
+
+RUN chown -R igvfd:igvfd /static/json/auditdoc.json
+
+RUN make-audit-docstring-json
```
keenangraham, line 77: I think run this after the full app has been copied over, then dump it in the /igvfd/src/static/json folder?

keenangraham (changes requested):
Couldn't we just "serve" this from Github?

I think if we do this we should probably just dump it into the existing /igvfd/src/static/ folder (https://github.com/IGVF-DACC/igvfd/blob/cbd653c35060cbd9cc20cc273dea22b670314117/src/igvfd/__init__.py#L42). Then you don't have to add a new route or deal with reading off disk.

ottojolanki:
> Couldn't we just "serve" this from Github?
> 
> I think if we do this we should probably just dump it into the existing /igvfd/src/static/ folder (
> 
> https://github.com/IGVF-DACC/igvfd/blob/cbd653c35060cbd9cc20cc273dea22b670314117/src/igvfd/__init__.py#L42
> 
> ). Then you don't have to add a new route or deal with reading off disk.

As I point out first thing in the ticket, you cannot do this because that dir gets masked by the volume in the docker compose. If we don't care about it not working locally, then we can do this.

docker/pyramid/Dockerfile
ottojolanki: That directory gets masked by docker compose volume.

docker/pyramid/Dockerfile
ottojolanki: See comment below. And on ticket.

src/igvfd/__init__.py
ottojolanki: Surely this won't get requested often enough for this to be even remotely an issue.

docker/pyramid/Dockerfile
ottojolanki: yep, pretty sure all of the audit doc stuff can go under that.

keenangraham:
> 

I think just selectively avoid mounting the static folder from local volume in all the Docker Compose yamls. We do this for node modules in frontend: https://github.com/IGVF-DACC/igvf-ui/blob/caea94bd3d683e765c3bf19f3d85eb219fef8c6d/docker-compose.yml#L11-L12

src/igvfd/__init__.py
keenangraham: I think it's worth getting into the static asset folder, otherwise this is just a dynamic view, not statically served.

ottojolanki:
> > 
> 
> I think just selectively avoid mounting the static folder from local volume in all the Docker Compose yamls. We do this for node modules in frontend: https://github.com/IGVF-DACC/igvf-ui/blob/caea94bd3d683e765c3bf19f3d85eb219fef8c6d/docker-compose.yml#L11-L12

K. This makes sense.

docker/pyramid/Dockerfile
ottojolanki: moved everything down.

src/igvfd/__init__.py
ottojolanki: the way to avoid getting masked was very helpful. doing it this way now. ty.

src/igvfd/commands/make_audit_docstring_json.py
ottojolanki: yup.

docker/pyramid/Dockerfile
ottojolanki: changed according to your advice below.

keenangraham (approved):
LGTM.

PR #445, opened 2023-07-14
IGVF-660-audit-curated-set-taxa

jenjou:


ian-whaling (changes requested):
Does it make sense to implement control for a mixed taxa sample, which would have no taxa in sample(s) but donors of multiple taxa?

jenjou:
> Does it make sense to implement control for a mixed taxa sample, which would have no taxa in sample(s) but donors of multiple taxa?

This should be handled now, so that if it's mixed donor and the sample taxa is none (which implies that it has mixed taxa) the audit doesn't trigger.

ian-whaling (approved):

ian-whaling (approved):

ian-whaling (dismissed):

ian-whaling (approved):

PR #452, opened 2023-07-20
IGVF-679-remove-seqspec

ian-whaling:


src/igvfd/tests/test_audit_measurement_set.py
```diff
@@ -142,7 +142,7 @@ def test_audit_inherit_related_multiome(
     testapp.patch_json(
         measurement_set_multiome['@id'],
         {
-            'seqspec': 'https://github.com/IGVF/seqspec/blob/main/assays/10x-ATAC/spec.yaml',
+            'protocol': 'https://www.protocols.io/view/example_protocol',
```
gabdank, line 145: why protocol is introduced here?

src/igvfd/tests/test_audit_measurement_set.py
```diff
@@ -154,41 +154,19 @@ def test_audit_inherit_related_multiome(
...
-        measurement_set['@id'],
-        {
-            'seqspec': 'https://github.com/IGVF/seqspec/blob/main/assays/10x-ATAC/spec.yaml'
-        }
-    )
-    res = testapp.get(measurement_set['@id'] + '@@audit')
-    assert all(
-        error['category'] != 'missing seqspec'
-        for error in res.json['audit'].get('WARNING', [])
+        error['category'] != 'missing protocol'
```
gabdank, line 168: I don't understand why protocol test is added in this PR?

src/igvfd/upgrade/file_set.py
```diff
@@ -61,3 +61,10 @@ def construct_library_2_3(value, system):
         value['origins'] = ['TF binding sites']
         notes += f' origins added via upgrade; update before removing note.'
         value['notes'] = notes.strip()
+
+
+@upgrade_step('measurement_set', '4', '5')
+def measurement_set_4_5(value, system):
+    # https://igvf.atlassian.net/browse/IGVF-679
+    if 'seqspec' in value:
+        del value['seqspec']
```
gabdank, line 70: you need to put seqspec info in the notes if you remove it

gabdank (changes requested):

src/igvfd/tests/test_audit_measurement_set.py
ian-whaling: This test was created to test that multiome measurement sets are inheriting the other's audits. The seqspec audit was used to test this, but since it has been removed in this PR I used the protocol audit instead, which is basically the same type of audit checking for a missing property.

src/igvfd/tests/test_audit_measurement_set.py
ian-whaling: see above

src/igvfd/upgrade/file_set.py
ian-whaling: added to notes

gabdank (dismissed):

gabdank (approved):

PR #457, opened 2023-07-21
IGVF-865-relax-assoc-disease-audit

iyoungworth:
update-mappings

### DIFF
--- a/src/igvfd/tests/test_audit_construct_library.py
+++ b/src/igvfd/tests/test_audit_construct_library.py
@@ -3,30 +3,10 @@ import pytest
 
 def test_audit_construct_library_associated_disease(
     testapp,
-    base_construct_library,
-    phenotype_term_alzheimers
+    base_construct_library
 ):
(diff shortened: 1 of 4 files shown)


src/igvfd/tests/test_audit_construct_library.py
```diff
@@ -6,27 +6,8 @@ def test_audit_construct_library_associated_disease(
     base_construct_library,
     phenotype_term_alzheimers
```
jenjou, line 7: I think phenotype_term_alzheimers can be dropped from this as it's no longer used in the test code.

jenjou (changes requested):
Minor change requested for the audit test. Looks good otherwise

src/igvfd/tests/test_audit_construct_library.py
iyoungworth: Thanks for catching! Removed

jenjou:
Looks OK but there is another conflict in the Mappings file.

jenjou (approved):

PR #459, opened 2023-07-21
IGVF-872-calculate-sources-in-multiplex

iyoungworth:


src/igvfd/upgrade/sample.py
```diff
@@ -234,3 +234,14 @@ def sample_12_13(value, system):
     # default value for 'virtual' property.
     # The default value will be automatically populated.
     return
+
+
+@upgrade_step('multiplexed_sample', '1', '2')
+def sample_13_14(value, system):
```
gabdank, line 240: i think we have to capture the popped values in "notes"

gabdank (changes requested):
record the deleted values in notes

src/igvfd/upgrade/sample.py
iyoungworth: Got it, I updated the upgrade

src/igvfd/tests/test_upgrade_multiplexed_sample.py
```diff
@@ -0,0 +1,10 @@
+import pytest
+
+
+def test_multiplexed_sample_upgrade_1_2(upgrader, multiplexed_sample_v1):
+    value = upgrader.upgrade('multiplexed_sample', multiplexed_sample_v1, current_version='1', target_version='2')
```
gabdank, line 5: please compare the notes

gabdank (changes requested):

src/igvfd/tests/test_upgrade_multiplexed_sample.py
iyoungworth: Done, added this test

gabdank (dismissed):

PR #467, opened 2023-07-26
IGVF-888-sequencefile-seqspec-audit

ian-whaling:


ian-whaling:
The latter audit for checking if the linked seqspec configurationfile has content_type "seqspec" can't be tested for since the enum for configuration files only has "seqspec" right now.

jenjou:
Was it specifically requested that these two audits share a single function?
I'm wondering if they should be split apart into a "missing seqspec" and "inconsistent seqspec" audit. It seems like we usually define the category as "missing ..." when a property is absent and I find that wording more intuitive as well, but if we are trying to reduce the number of categories I can understand merging them.

jenjou (approved):

gabdank (approved):

PR #471, opened 2023-07-28
IGVF-890-support-multiple-types-in-report-view

mingjiecn:


src/igvfd/search_views.py
```diff
@@ -96,14 +99,15 @@ def search_generator(request):
...
-            )
+            ),
         ]
     )
     return fgr.render()
 
 
 @view_config(route_name='report', request_method='GET', permission='search')
 def report(context, request):
+    validate_report(request)
```
keenangraham, line 110: I would like to not block this view with type validation if possible. It's basically a search view which doesn't care. Can we just validate in the view that actually generates the TSV?

src/igvfd/search_views.py
```diff
@@ -258,3 +262,36 @@ def top_hits(context, request):
...
+            is_logic_group = True
+            for subtype in subtypes_to_query:
+                if subtype not in subtypes:
+                    is_logic_group = False
+                    break
+            if is_logic_group:
+                break
+        if not is_logic_group:
+            msg = f'Report view does not support the group of types: {types_to_query}'
+            raise HTTPBadRequest(explanation=msg)
```
keenangraham, line 297: Can we move this into same file as `report_download`?

keenangraham (changes requested):

keenangraham:
The logic seems pretty complicated. What happened to the idea of using only the types the exist in the search results, instead of the types specified in the query string?

src/igvfd/search_views.py
mingjiecn: @gabdank  wants the logic implemented at both report view and report download view so the frontend won’t need to display the report view for types report that we don't support for downloading.

mingjiecn:
> The logic seems pretty complicated. What happened to the idea of using only the types that exist in the search results, instead of the types specified in the query string?

The logic in the validation make sure the types we search are from the same parent(Item excluded). We don't want to support types that can't logically grouped together.

src/igvfd/search_views.py
gabdank: I think the same logic should be applicable for both: report-view and report-download. 
From talking to Forrest it sounds that the front-end at this point relies on the backend to supply the right answer - for both and I am not sure we should change it.

gabdank:
`What happened to the idea of using only the types that exist in the search results, instead of the types specified in the query string?`

search query "type=Sample&type=FileSet" will include types with columns that I think should not be in the same table

PR #474, opened 2023-07-28
spin-up-new-opensearch

keenangraham:


ottojolanki (approved):
Confirmed that for all envs:
New instance type: `m6g.xlarge.search`
engine version: `EngineVersion.OPENSEARCH_2_7`
volume_size: `30`
and we are writing to the new `'Opensearch27'`

PR #475, opened 2023-07-28
IGVF-927-upgrade-lambda-python-runtime

ottojolanki:


keenangraham (approved):

keenangraham (changes requested):
This probably can't be merged until we are doing full database refresh of lower environments. Suspect it will trigger snapshot function again, and existing databases will get conflicting snapshot.

PR #484, opened 2023-07-31
IGVF-895-pluralize-3-properties

iyoungworth:


src/igvfd/upgrade/sample.py
```diff
@@ -192,10 +192,10 @@ def whole_organism_10_11(value, system):
 def whole_organism_11_12(value, system):
```
gabdank, line 192: I am not 100% sure this upgrade change is "kosher" or even going to work, as in schema 11 there was no sample_terms there?

src/igvfd/tests/test_upgrade_whole_organism.py
```diff
@@ -89,7 +89,7 @@ def test_whole_organism_upgrade_10_11(upgrader, whole_organism_v10):
 def test_whole_organism_upgrade_11_12(upgrader, whole_organism_v11):
     value = upgrader.upgrade('whole_organism', whole_organism_v11, current_version='11', target_version='12')
     assert value['schema_version'] == '12'
-    assert value['biosample_term'] == '/sample-terms/UBERON_0000468/'
+    assert value['sample_terms'] == ['/sample-terms/UBERON_0000468/']
```
gabdank, line 92: if the fixture is not updated and this test is not updated - anything fails?

gabdank (commented):
questions about old update of whole organism

src/igvfd/upgrade/sample.py
iyoungworth: I think it is necessary, the upgrade fails if I don't make this change (I think the issue is that even though v12 didn't exist at the time, there isn't a stored version WholeOrganism schema_version:11 to run the upgrade with)
>       if value['biosample_term'] != ['/sample-terms/UBERON_0000468/']:
E       KeyError: 'biosample_term'

src/igvfd/tests/test_upgrade_whole_organism.py
iyoungworth: The test is updated to match the upgrade change

jenjou (changes requested):

jenjou (approved):

src/igvfd/upgrade/donor.py
```diff
@@ -142,3 +142,12 @@ def donor_10_11(value, system):
...
         value['publication_identifiers'] = value['references']
         del value['references']
+
+
+@upgrade_step('rodent_donor', '9', '10')
+def rodent_donor_9_10(value, system):
+    # https://igvf.atlassian.net/browse/IGVF-895
+    # Source property is pluralized
+    if 'source' in value:
+        value['sources'] = value['source']
```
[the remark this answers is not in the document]
iyoungworth: Thank you for catching this! Inadvisable copy-paste from an old upgrade

PR #490, opened 2023-08-02
IGVF-890-report

mingjiecn:


src/igvfd/search_views.py
```diff
@@ -117,7 +117,7 @@ def report(context, request):
             ),
             IDResponseField(),
             ContextResponseField(),
-            BasicReportWithFacetsResponseField(
+            MultipleTypesReportWithFacetsResponseField(
```
keenangraham, line 120: Make a new report endpoint instead of changing this one. Something like `multitype-report` or `multireport`?

src/igvfd/report.py
```diff
@@ -56,20 +62,37 @@ def _convert_camel_to_snake(type_str):
...
+        },
+        response_fields=[
+            MultipleTypesReportWithFacetsResponseField(
+                default_item_types=DEFAULT_ITEM_TYPES,
+                reserved_keys=RESERVED_KEYS,
+            ),
+            ColumnsResponseField(),
+
+        ]
+    )
```
keenangraham, line 78: We should be calling the existing view instead of recreating it here. Something like `response = request.embed(f'/multitype-report/{original_query_string_with_limit_0_added}')`

src/igvfd/report.py
```diff
@@ -56,20 +62,37 @@ def _convert_camel_to_snake(type_str):
...
+    facets = response.json['facets']
+    abstract_types = get_abstract_types(request)
+    types_in_search_result = []
+    for facet in facets:
+        if facet['field'] == 'type':
+            for term in facet['terms']:
+                type_name = term['key']
+                if type_name not in abstract_types:
+                    types_in_search_result.append(term['key'])
+            break
```
keenangraham, line 89: What is the `break` doing?

src/igvfd/tests/indexing/test_report_tsv.py
```diff
@@ -26,6 +26,26 @@ def test_batch_download_report_download(workbook, testapp):
         b'Contact P.I.', b'Title'
     ]
 
+    res = testapp.get('/report.tsv?type=Award&field=contact_pi&field=title&config=Award')
```
keenangraham, line 29: Make new tests specifically for `multitype-report`.

src/igvfd/report.py
```diff
@@ -56,20 +62,37 @@ def _convert_camel_to_snake(type_str):
 @view_config(route_name='report_download', request_method='GET')
 def report_download(context, request):
```
keenangraham, line 63: Leave this endpoint alone and make new endpoint e.g. `multitype_report_download`.

README.md
```diff
@@ -56,6 +56,26 @@ $ pytest
 docker compose down -v
 ```
 
+Or run indexer tests interactively:
```
keenangraham, line 59: Would it be cleaner to put this in its own PR?

keenangraham (changes requested):

src/igvfd/report.py
mingjiecn: break after find facet "type" and get all the term keys.

src/igvfd/search_views.py
mingjiecn: Sounds good, will do.

src/igvfd/report.py
mingjiecn: Got it. Thanks!

PR #492, opened 2023-08-03
IGVF-916-add-zim3-to-fused-domain

KAndreeva:
Added ZIM3 to fused domain enums in modification.

### DIFF
--- a/src/igvfd/mappings/modification.json
+++ b/src/igvfd/mappings/modification.json
@@ -1,6 +1,6 @@
 {
-    "hash": "40f53d6047fe482a4bae05546d03c7b0",
-    "index_name": "modification_40f53d60",
+    "hash": "e878415f9709f8a7699cfe0c372cb12a",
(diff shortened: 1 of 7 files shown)


jenjou (changes requested):
Code change looks fine. However, I would suggest removing the new test in test_schema_modification, and instead adding a new insert using the new enum value in this file: https://github.com/IGVF-DACC/igvfd/blob/dev/src/igvfd/tests/data/inserts/modification.json

As long as this insert can be loaded, that is usually sufficient to confirm the enum is working correctly and we don't need to add another test function.

jenjou:
Looks good but two comments:
1. PR needs rebase (again).
2. Since we have modified the "summary" calculation I think we can add another case for the test in test_types_modification.py to make sure that we are correctly avoiding the "SpSpCas" situation.

jenjou (approved):

PR #495, opened 2023-08-04
IGVF-930-add-readout-to-measurement

iyoungworth:


jenjou (approved):

jenjou (commented):
Code looks OK but there are merge conflicts.
I was wondering if there was a way to combine the 2 audits together since most of the code is the same; but that would likely cause issues with the docstrings. I think it's okay as is unless we have a clever way to merge it into 1 audit.

iyoungworth:
Yes, the docstring is why I kept it separate- I think it's hypothetically possible to merge them but it would be less clear. Let me know what you think, eg:
**audit_detail**: MeasurementSet IGVFSxxx has an inconsistent readout property. Screening assays (such as CRISPR screen and MPRA) are required to specify a readout, other assays should not include one.
**audit_category**: inconsistent readout

jenjou (approved):

PR #504, opened 2023-08-09
IGVF-580-inconsistent-libraries-audit

ian-whaling:


src/igvfd/audit/measurement_set.py
```diff
@@ -79,7 +79,38 @@ def audit_unspecified_protocol(value, system):
...
             f'the assay on protocols.io.'
         )
         yield AuditFailure('missing protocol', detail, level='NOT_COMPLIANT')
+
+
+@audit_checker('MeasurementSet', frame='object')
+def audit_construct_libraries(value, system):
+    '''
+        audit_detail: Construct libraries linked to in a measurement set are expected to have the same library details.
+        audit_category: inconsistent construct library metadata
```
iyoungworth, line 92: Do you feel strongly about 'metadata'? Swapping in 'details' might be nice because it matches the property names being compared

iyoungworth (dismissed):
Looks great- 1 very minor word choice question for the audit category, but no necessary changes

src/igvfd/audit/measurement_set.py
ian-whaling: I agree with this change.

PR #508, opened 2023-08-11
IGVF-592-add-matrix-file

meenakshikagda:


src/igvfd/schemas/matrix_file.json
```diff
@@ -0,0 +1,107 @@
...
+    },
+    "type": "object",
+    "properties": {
+        "schema_version": {
+            "default": "1"
+        },
+        "content_type": {
+            "comment": "Content Type describes the content of the file.",
+            "enum": [
+                "sparse gene count matrix",
```
jenjou, line 58: Small detail, but can we order the enum lists (including the dimension1 and 2) to be alphabetical? I find it easier to read later on as we add more enums.

src/igvfd/schemas/matrix_file.json
```diff
@@ -0,0 +1,107 @@
...
+                "treatment",
+                "time",
+                "variant"
+            ]
+        },
+        "file_format": {
+            "enum": [
+                "h5ad",
+                "hdf5",
+                "h5ad",
```
jenjou, line 93: h5ad is listed twice here

src/igvfd/schemas/matrix_file.json
```diff
@@ -0,0 +1,107 @@
...
+                "matrixmarket"
```
jenjou, line 94: I think the file extension for this is "mtx" so maybe that should be the enum? We may need to double check.

cricketsloan: We certainly have .mtx.gz files coming from Cell ranger. I also have seen this come in a tar of 3 files.  Specifically these 3 https://support.10xgenomics.com/single-cell-multiome-atac-gex/software/pipelines/latest/output/matrices I am not sure if we are going to accept them merged like that.

cricketsloan (commented):
I think jennifer has the few points that need addressing.  I have added a comment about .mtx files

jenjou (approved):

PR #519, opened 2023-08-16
IGVF-987-audit-error-construct-libraries

ian-whaling:


iyoungworth (dismissed):
Seems like this should work- QA will have to double check that it "reads" well. Test covers this case so at least audit script error seems to be avoided!

jenjou (approved):

PR #527, opened 2023-08-22
IGVF-1009-seqspec-reverse-link

ian-whaling:


src/igvfd/types/file.py
```diff
@@ -409,6 +415,18 @@ def unique_keys(self, properties):
                 keys.setdefault('alias', []).append(value)
         return keys
 
+    @calculated_property(schema={
+        'title': 'Sequence files this file is a seqspec of',
```
jenjou, line 419: I think the title should be adjusted. At minimum, it should be in Title Case, but also wondering if we can reword to to be more concise. I think "Sequence files this file is a seqspec of" could be kept as the "description".

Would it be confusing if the title was just "Seqspec Of"?

ian-whaling: I based the format off a similar property https://github.com/IGVF-DACC/igvfd/blob/dev/src/igvfd/types/file_set.py#L55-L65, but I'll adjust this.

jenjou (dismissed):

gabdank (approved):

PR #548, opened 2023-08-30
IGVF-1053-modification-id-properties

khzlin:


iyoungworth (changes requested):
Can you please also remove these fields from the columns in searches/configs file?

iyoungworth (approved):

gabdank (approved):

PR #549, opened 2023-08-31
IGVF-1025-update-sample-summary

ian-whaling:


src/igvfd/tests/test_types_tissue.py
```diff
@@ -0,0 +1,61 @@
...
+    testapp.patch_json(
+        tissue['@id'],
+        {
+            'donors': [rodent_donor['@id'], parent_rodent_donor_2['@id']],
+            'treatments': [treatment_chemical['@id']]
+        }
+    )
+    res = testapp.get(tissue['@id'])
+    assert res.json.get(
+        'summary') == f'virtual embryonic brown adipose tissue, mixed sex Mus musculus strain1, strain3 (50-100 days) {treatment_summary}'
```
gabdank, line 61: i think the assertion should include the treatment summary written down, not parametrized. if I understand correctly the treatment summary will be capitalized and will appear "disconnected" from the tissue summary. but I am not sure that is true.

src/igvfd/tests/test_types_whole_organism.py
```diff
@@ -0,0 +1,45 @@
...
+        {
+            'donors': [human_donor['@id'], rodent_donor['@id']],
+            'lower_bound_age': 10,
+            'upper_bound_age': 10,
+            'age_units': 'month'
+        }
+    )
+    print(res.json.get('summary'))
+    res = testapp.get(whole_organism['@id'])
+    assert res.json.get('summary') == '2 whole organisms, males (10 months)'
```
gabdank, line 45: i think rodent/human should say mixed taxa?

src/igvfd/types/sample.py
```diff
@@ -200,6 +177,98 @@ def taxa(self, request, donors):
         if len(taxas) == 1:
             return list(taxas).pop()
 
+    @calculated_property(
```
gabdank, line 180: would not it make sense to have individual sample types of summaries, instead of putting them all together in one big function?

src/igvfd/types/sample.py
```diff
@@ -200,6 +177,98 @@ def taxa(self, request, donors):
...
+                    sex = f'{sex}s'
+                summary_terms = f'{summary_terms} {sex}'
+        if taxa and self.item_type in ['primary_cell', 'in_vitro_system', 'tissue', 'whole_organism']:
+            if taxa == 'Mus musculus':
+                strains = []
+                for donor in donors:
+                    donor_object = request.embed(donor, '@@object?skip_calculated=true')
+                    strains.append(donor_object['strain'])
+                strains = ', '.join(sorted(list(set(strains))))
+                taxa = f'{taxa} {strains}'
```
gabdank, line 239: i am not clear what happens if there is mix of taxas?

gabdank (changes requested):

src/igvfd/types/sample.py
ian-whaling: No strains are listed if taxa is mixed.

src/igvfd/types/sample.py
ian-whaling: In the commit [in vitro system summary update](https://github.com/IGVF-DACC/igvfd/pull/549/commits/4918aed7e84b8900c23e47884958f67f51a00e3c) I had them listed in individual summaries, so if you take a look and really prefer it I can change it back. 

But I really don't think this is better since there is so much redundant code between the summaries and if there were to be further edits to the summaries they would be easily modified here whereas the separate summaries would all need to all be edited with the same changes. Although the summary is longer, the child summaries are just as long, I think this is cleaner.

src/igvfd/tests/test_types_whole_organism.py
ian-whaling: If this is desired why doesn't mixed taxa return anything in the calculated property for taxa?

src/igvfd/tests/test_types_tissue.py
ian-whaling: I'll change this.

gabdank: ok

src/igvfd/tests/test_types_whole_organism.py
gabdank: it is not calculated, because it is mixed. the summary however could "tap" into donors to receive the taxa.  I am trying to say that strict property listing taxa - should avoid saying "mixed" as it is not a taxa. the summary which is human human-readable string can say whatever makes sense. in this case maybe you can say "2 whole organisms, homo sapiens and mus musculus, males (10 months)" or something along these lines. it seems wrong not to say the taxa in the summary if they are available in donors.

src/igvfd/types/sample.py
ian-whaling: I can add strains if mixed taxa includes mice, I excluded since I assumed taxa would be excluded if it were mixed.

src/igvfd/tests/test_types_whole_organism.py
ian-whaling: Ok. I'll change this.

gabdank (dismissed):

src/igvfd/tests/test_types_multiplexed_sample.py
```diff
@@ -0,0 +1,20 @@
...
+        multiplexed_sample['@id'],
+        {
+            'multiplexed_samples': [tissue['@id'], in_vitro_cell_line['@id'], whole_organism['@id']],
+        }
+    )
+    whole_organism_summary = testapp.get(whole_organism['@id']).json.get('summary')
+    res = testapp.get(multiplexed_sample['@id'])
+    assert res.json.get('summary').endswith(', ... and 1 more sample')
+    assert sum(1 for summary in [tissue_summary, in_vitro_cell_line_summary,
+               whole_organism_summary] if summary in res.json.get('summary')) == 2
```
gabdank, line 20: not sure about this assertion,why not compare the summary as it is without parameters and looping through it?

src/igvfd/tests/test_types_multiplexed_sample.py
```diff
@@ -0,0 +1,20 @@
...
+        'summary') and tissue_summary in res.json.get('summary')
+    testapp.patch_json(
+        multiplexed_sample['@id'],
+        {
+            'multiplexed_samples': [tissue['@id'], in_vitro_cell_line['@id'], whole_organism['@id']],
+        }
+    )
+    whole_organism_summary = testapp.get(whole_organism['@id']).json.get('summary')
+    res = testapp.get(multiplexed_sample['@id'])
+    assert res.json.get('summary').endswith(', ... and 1 more sample')
```
gabdank, line 18: please compare the whole summary

gabdank (changes requested):
assertion comment

gabdank (dismissed):

PR #556, opened 2023-09-05
IGVF-1037-add-construct-library-set

iyoungworth:


src/igvfd/schemas/measurement_set.json
```diff
@@ -122,6 +123,7 @@
             "linkTo": "AssayTerm"
         },
         "moi": {
+            "permission": "import_items",
```
gabdank, line 126: add to the comment indication we are going to move the property

gabdank (changes requested):
minor comments

gabdank (approved):

PR #559, opened 2023-09-07
IGVF-1062-matrix-file-facets

ian-whaling:


src/igvfd/searches/configs/matrix_file.py
```diff
@@ -6,6 +6,67 @@
 )
 def matrix_file():
     return {
+        'facets': {
+            'uuid': {
+                'title': 'UUID'
```
jenjou, line 11: UUID and accession don't seem to be included in facets for other File subclasses except Signal File. I'm confused if we should be adding UUID and accession to every subclass, or if we should be omitting them?

src/igvfd/searches/configs/matrix_file.py
```diff
@@ -25,6 +86,12 @@ def matrix_file():
             'lab': {
                 'title': 'Lab'
             },
+            'file_set': {
+                'title': 'File Set'
+            },
+            'content_summary': {
+                'title': 'Content Summary'
+            },
             'status': {
```
jenjou, line 95: Should upload_status also be added to columns?

src/igvfd/searches/configs/matrix_file.py
ian-whaling: Yes I coped these from signal file, thanks for catching that I don't see the use in having uuid and accession in the facets. It looks like Mingjie had added them there but I don't think they serve any purpose on the search page so I will remove them.

src/igvfd/searches/configs/matrix_file.py
ian-whaling: upload_status is already in the columns

jenjou (approved):

PR #568, opened 2023-09-12
IGVF-1090-avoid-empty-array-selection

iyoungworth:


src/igvfd/types/file_set.py
```diff
@@ -461,6 +461,8 @@ def summary(self, request, file_set_type, scope, selection_criteria, genes=None,
         pheno_terms = []
         pheno_phrase = ''
         preposition = ''
+        criteria = []
+        criteria = criteria + selection_criteria
```
iyoungworth, line 466: Somehow, if I only set criteria = selection_criteria and all other changes are the same, selection_criteria on the object STILL gets edited. This workaround looks like it works (see tests, and could confirm insert on local demo still had the property populated)

forresttanaka: Yeah, that makes sense, since `criteria` would just be a reference to the same memory pointed to be `selection_criteria`.

forresttanaka (approved):
😎

PR #571, opened 2023-09-18
IGVF-1092-assay-title-enum

khzlin:


jenjou (changes requested):
Can we add one more insert for snMCT-seq enum?
Otherwise code is OK.

jenjou (approved):

PR #577, opened 2023-09-20
IGVF-1111-samples-donors-dependency

ian-whaling:


src/igvfd/types/file_set.py
```diff
@@ -288,6 +310,29 @@ def summary(self, request, assay_term, preferred_assay_title=None, readout=None,
                 sentence += phrase
         return sentence
 
+    @calculated_property(
+        condition='samples',
+        schema={
+            'title': 'Donors',
+            'description': 'The donors of the samples associated with this measurement set.',
```
gabdank, line 317: curated sets rather than measurment

src/igvfd/types/file_set.py
```diff
@@ -385,6 +430,29 @@ def summary(self, request, file_set_type, measurement_sets=None):
...
+            'type': 'array',
+            'uniqueItems': True,
+            'items': {
+                'title': 'Donor',
+                'description': 'Donor of a sample associated with this measurement set.',
+                'type': 'string',
```
gabdank, line 443: auxiliary sets rather than measurement

src/igvfd/types/file_set.py
```diff
@@ -104,6 +103,29 @@ def assay_title(self, request, input_file_sets=None):
...
+                'linkTo': 'Donor'
```
gabdank, line 117: analysis sets rather than measurement?

gabdank (changes requested):
I think comments in the donor calculation were copied form measurment set, and were supposed to be class specific. I also think that the function itself is duplicated 3 times - so maybe we should refactor and have it called 3 times, rather than have duplicated code

src/igvfd/types/file_set.py
ian-whaling: this is for measurement sets. curated sets do not calculate donors, the property is still submittable for them.

src/igvfd/types/file_set.py
ian-whaling: fixed

src/igvfd/types/file_set.py
ian-whaling: fixed

src/igvfd/schemas/prediction_set.json
```diff
@@ -4,10 +4,23 @@
     "$id": "/profiles/prediction_set.json",
     "$schema": "https://json-schema.org/draft/2020-12/schema",
     "type": "object",
-    "required": [
-        "lab",
-        "award",
-        "file_set_type"
+    "anyOf": [
```
gabdank, line 7: I think it should be oneOf

gabdank (changes requested):
I thin it should be oneOf and not anyOf

gabdank (approved):

PR #584, opened 2023-09-25
IGVF-1122-add-result-columns-to-multireport

mingjiecn:


src/igvfd/tests/indexing/test_searches_fields.py
```diff
@@ -0,0 +1,16 @@
...
+    rrf = ResultColumnsResponseField()
+    assert isinstance(rrf, ResultColumnsResponseField)
+
+
+def test_searches_fields_result_columns_response_field_query(workbook, testapp):
+    r = testapp.get(
+        '/multireport/?institute_label=Stanford'
+    )
+    assert 'result_columns' in r.json
+    assert len(r.json['result_columns']) < len(r.json['columns'])
```
keenangraham, line 16: Can we also have a test that expects some specific columns just to make sure we are getting what we want?

src/igvfd/searches/fields.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def get_abstract_types(request):
+    types = []
+    item_registry = request.registry[TYPES]
+    for name, item in item_registry.abstract.items():
+        subtypes = item.subtypes
+        if len(subtypes) > 1:
+            types.append(name)
+    return types
```
keenangraham, line 88: Can we keep this stuff in `report.py` for now and import it here? These deserve more clean up and refactoring if we move, and think `fields.py` should only have `Fields`.

keenangraham (changes requested):
Looks good, just a couple things.

src/igvfd/searches/fields.py
mingjiecn: Put those functions in report.py or search_views.py will cause circular import so I have to put them here in fields.py

src/igvfd/tests/indexing/test_searches_fields.py
mingjiecn: sure

src/igvfd/searches/fields.py
keenangraham: Still need better solution for this.

src/igvfd/tests/indexing/test_searches_fields.py
```diff
@@ -0,0 +1,18 @@
...
+    assert 'pi' in r.json['result_columns']
+    assert 'institute_label' in r.json['result_columns']
```
keenangraham, line 18: Why not test for all of the fields that are expected?

keenangraham (changes requested):

src/igvfd/report.py
```diff
@@ -5,7 +5,7 @@
 from snovault.elasticsearch.searches.interfaces import SEARCH_CONFIG
 from snosearch.parsers import QueryString
 from snovault.compat import bytes_
-from igvfd.search_views import search_generator
+from igvfd.search_generator import search_generator
```
keenangraham, line 8: Looks good but can we put this in `/searches/generator.py`?

keenangraham (approved):

keenangraham (approved):
LGTM.

PR #585, opened 2023-09-25
IGVF-1002-fileset-search-config

forresttanaka:
This mostly involves adding a search config to abstract types, and adding a `type` property to these search configs as well as their child search configs.

I removed the `assembly` and `transcriptome_annotation` properties from the AlignmentFile search config because these properties don’t exist in AlignmentFile objects anymore.


forresttanaka:
I notice I haven’t updated all the type-facet titles. Closing the PR and will reopen after I fix those.

PR #626, opened 2023-10-12
IGVF-1140-reverse-link-on-parents

ian-whaling:


src/igvfd/types/sample.py
```diff
@@ -321,6 +351,28 @@ def summary(self, request, sample_terms, donors, sex, age, age_units=None, embry
...
+        return paths_filtered_by_status(request, parent_of)
+
+    @calculated_property(schema={
+        'title': 'Pooled In',
+        'type': ['string', 'object'],
+        'linkFrom': 'Sample.pooled_from',
+        'notSubmittable': True,
+    })
+    def pooled_in(self, request, pooled_in):
+        return paths_filtered_by_status(request, pooled_in)
```
ian-whaling, line 374: This returns an array when I'd like it to just be a string since it will always just be a single link. Not sure how to accomplish this.

ian-whaling: Never mind, an array is fine since a sample may be pooled in multiple other samples.

src/igvfd/types/sample.py
```diff
@@ -321,6 +351,28 @@ def summary(self, request, sample_terms, donors, sex, age, age_units=None, embry
 
         return summary_terms.strip(',')
 
+    @calculated_property(schema={
+        'title': 'Parted Samples',
```
gabdank, line 355: i think "Biosample Parts" would be better than "Parted Samples"

src/igvfd/types/sample.py
```diff
@@ -321,6 +351,28 @@ def summary(self, request, sample_terms, donors, sex, age, age_units=None, embry
...
+        'type': 'array',
+        'items': {
+            'title': 'Parted Child Sample',
```
gabdank, line 358: I think "Biosample Part" would be better than "Parted Child Sample"

src/igvfd/types/sample.py
```diff
@@ -90,6 +92,32 @@ def file_sets(self, request, file_sets):
...
+        'notSubmittable': True,
+    })
+    def sorted_fractions(self, request, sorted_fractions):
+        return paths_filtered_by_status(request, sorted_fractions)
+
+    @calculated_property(schema={
+        'title': 'Origin Sample Of',
+        'type': 'array',
+        'items': {
+            'title': 'Child Sample',
```
gabdank, line 112: `Originated Sample` would be better than "Child sample"

gabdank (changes requested):

gabdank (approved):

PR #631, opened 2023-10-17
IGVF-1051-fix-crispr-audit

ian-whaling:


src/igvfd/audit/measurement_set.py
```diff
@@ -96,8 +96,9 @@ def audit_inconsistent_readout(value, system):
     assay = system.get('request').embed(assay_term, '@@object?skip_calculated=true')
     screen_assays = ['Perturb-seq',
                      'CRISPR screen',
-                     'massively parallel reporter assay']
-    if assay.get('term_name') in screen_assays:
+                     'massively parallel reporter assay',
+                     'cas mediated mutagenesis']
+    if assay.get('term_name') in screen_assays or value.get('preferred_assay_title', '') in screen_assays:
```
gabdank, line 101: MPRA is not a screen assay, i think we need to rename the list to be "assays_with_readout" or something like that

src/igvfd/audit/measurement_set.py
```diff
@@ -134,25 +135,22 @@ def audit_inconsistent_modifications(value, system):
         yield AuditFailure('inconsistent modifications', detail, level='NOT_COMPLIANT')
 
 
-@audit_checker('MeasurementSet', frame='object')
+@audit_checker('MeasurementSet', frame='embedded')
```
gabdank, line 138: we should not rely on embedding in my opinion

gabdank (changes requested):

src/igvfd/audit/measurement_set.py
ian-whaling: reverted to object

src/igvfd/audit/measurement_set.py
ian-whaling: reworded

gabdank (approved):

gabdank (approved):

PR #647, opened 2023-10-25
IGVF-862-dbxref-reference-files

ian-whaling:


src/igvfd/upgrade/file.py
```diff
@@ -57,3 +57,13 @@ def sequence_file_3_4(value, system):
...
         value['notes'] = notes
+
+
+@upgrade_step('reference_file', '4', '5')
+@upgrade_step('matrix_file', '1', '2')
+@upgrade_step('signal_file', '1', '2')
+@upgrade_step('configuration_file', '1', '2')
+@upgrade_step('alignment_file', '1', '2')
+def file_4_5(value, system):
+    if 'dbxrefs' in value:
```
gabdank, line 68: should not this step be true ONLY if dbxref is an empty string?

gabdank (changes requested):

src/igvfd/upgrade/file.py
ian-whaling: these files atp can only have dbxrefs if it consists of empty string(s)

gabdank (dismissed):

src/igvfd/upgrade/file.py
```diff
@@ -57,3 +57,18 @@ def sequence_file_3_4(value, system):
         else:
             notes = f'The mean read length previously exceeded the upper limit of 300000000 so was set to 300000000, please assign the appropriate length.'
         value['notes'] = notes
+
+
+@upgrade_step('reference_file', '4', '5')
+@upgrade_step('matrix_file', '1', '2')
+@upgrade_step('signal_file', '1', '2')
+@upgrade_step('configuration_file', '1', '2')
+@upgrade_step('alignment_file', '1', '2')
```
gabdank, line 66: i think the schema versions here are incorrect, for example sequence_file is 4->5

gabdank (changes requested):

src/igvfd/upgrade/file.py
ian-whaling: fixed

src/igvfd/schemas/changelogs/file.md
```diff
@@ -1,5 +1,4 @@
 ## Changelog for *`file.json`*
 
-### Release v1
-
-* Deprecate `reference_data` and `sequence_data` schemas.
+* Disallow submission of empty string in `dbxrefs`. (11/01/2023)
```
iyoungworth, line 3: Should update this to next version date

iyoungworth (changes requested):
General request for concrete file schemas: if minItems restriction is new to all cases, I think each changelog should mention that too

For file schemas without their own dbxrefs pattern (all but Reference and Sequence): please add a dbxrefs section with comment indicating it isn't currently submittable (a specific pattern isn't present, so nothing will match File regex pattern "(?!)"

src/igvfd/schemas/changelogs/file.md
ian-whaling: updated

src/igvfd/schemas/genome_browser_annotation_file.json
```diff
@@ -170,6 +170,9 @@
                 "appscript": "GRCh38",
                 "igvf_utils": "GRCh38"
             }
+        },
```
iyoungworth, line 173: The schema version needs to be bumped

src/igvfd/schemas/tabular_file.json
```diff
@@ -178,6 +178,9 @@
                 "appscript": "GRCh38",
                 "igvf_utils": "GRCh38"
             }
+        },
```
iyoungworth, line 181: The schema version needs to be bumped up

iyoungworth (changes requested):
New dbxrefs comments and changelogs look great, but the 2 new concrete File classes need default schema version bumped to '2'

iyoungworth (approved):

PR #661, opened 2023-11-03
IGVF-1243-add-full-download-url-to-report

mingjiecn:
There types: File, Document and Image has href. For Document and Image, href is embedded in attachment. What I did in this ticket is to complete the href to make a full length URL in downloaded report.


keenangraham (changes requested):
Need some tests to ensure the different cases are working as expected and protect the feature from breaking in the future.

keenangraham (approved):

PR #665, opened 2023-11-08
IGVF-1257-embed-observation-date

ian-whaling:


src/igvfd/types/donor.py
```diff
@@ -26,7 +26,8 @@ class Donor(Item):
         Path('award', include=['@id', 'component']),
         Path('lab', include=['@id', 'title']),
         Path('submitted_by', include=['@id', 'title']),
-        Path('phenotypic_features.feature', include=['@id', 'feature', 'term_name', 'quantity', 'quantity_units'])
+        Path('phenotypic_features.feature', include=['@id', 'feature',
```
khzlin, line 29: phenotypic_features.feature.quantity + quantity_units + observation_date doesn't exist. Shouldn't this be separated into two different lines?

`Path('phenotypic_features.feature' ...)`
`Path('phenotypic_features' ...)`

khzlin (changes requested):

src/igvfd/types/donor.py
ian-whaling: You are correct that these properties do not exist for phenotypic_features.feature, but the embedding function works to embed the properties from all objects it follows up to the top object. This is how I've been advised by Otto and Idan to do it since it avoids traversing the phenotypic_features object twice as how it would with separating it into two lines.

khzlin: oh i was unaware. If that's the case, that makes sense.

khzlin (approved):

gabdank (approved):

PR #685, opened 2023-11-16
IGVF-1171-meas-set-type

jenjou:


src/igvfd/schemas/measurement_set.json
```diff
@@ -197,6 +198,16 @@
                 "appscript": "[\"john-doe:auxiliary_set01\", \"TSTDS98084731\"]",
                 "igvf_utils": "john-doe:auxiliary_set01, TSTDS98084731"
             }
+        },
```
iyoungworth, line 201: Do we want to consider using a default value here to make submission simpler, since we only have 1 enum available?

iyoungworth (commented):
I think since only 1 enum is available on MeasurementSet file_set_type, we might assign a default value that would be provided automatically on submission (we have this on WholeOrganism for sample_terms). The potential downside is that if we have additional enum in the future, that would be the point at which submitters would need to be aware of the needed to use the property

iyoungworth (approved):

iyoungworth (approved):

gabdank (approved):

PR #686, opened 2023-11-17
IGVF-1021-mapping-nic-audit

jenjou:


ian-whaling (changes requested):
I think we should use term_name instead.

src/igvfd/audit/biosample.py
```diff
@@ -16,12 +16,38 @@ def audit_biosample_nih_institutional_certification(value, system):
...
-        )
-        yield AuditFailure('missing nih_institutional_certification', detail, level='ERROR')
+        # Only audit if the sample is associated with at least one non-characterization assay.
+        assay_titles = []
+        assay_ids = []
+        for fs in value.get('file_sets', []):
+            fs_object = system.get('request').embed(fs + '@@object?skip_calculated=true')
+            assay_titles.append(fs_object.get('preferred_assay_title', ''))
+            fs_assay_term = system.get('request').embed(fs_object['assay_term'] + '@@object?skip_calculated=true')
+            assay_ids.append(fs_assay_term['term_id'])
```
ian-whaling, line 26: I think we should use term_name instead.

ian-whaling (changes requested):
I think we should use term_name instead.

PR #690, opened 2023-11-22
IGVF-1163-fix-regex-model-set

ian-whaling:


src/igvfd/upgrade/software_version.py
```diff
@@ -1,6 +1,23 @@
 from snovault import upgrade_step
 
 
+@upgrade_step('software_version', '2', '3')
```
gabdank, line 4: should not this code be below the 1-2 update?

the upgrade seems to be getting rid of the version in some cases, but it is a required property?

gabdank (changes requested):

src/igvfd/upgrade/software_version.py
ian-whaling: Fixed & fixed.

gabdank (approved):

PR #694, opened 2023-11-28
IGVF-1064-fix-audit-array

khzlin:


gabdank (changes requested):
please add a test that confirms proper construction of the audit detail

gabdank (approved):

PR #701, opened 2023-11-30
IGVF-1299-image-file

ian-whaling:


src/igvfd/schemas/image_file.json
```diff
@@ -0,0 +1,86 @@
+{
+    "title": "Image File",
+    "$id": "/profiles/image_file.json",
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "description": "A file containing image data in jpg or png format.",
```
gabdank, line 5: I think the description should not limit to specific file formats.

src/igvfd/schemas/image_file.json
```diff
@@ -0,0 +1,86 @@
...
+    "dependentSchemas": {
+        "$merge": [
+            "file.json#/dependentSchemas"
+        ]
+    },
+    "type": "object",
+    "properties": {
+        "schema_version": {
+            "default": "2"
+        },
```
gabdank, line 48: why 2?

ian-whaling: left over from copying another schema. I've set it back to 1.

src/igvfd/schemas/image_file.json
ian-whaling: fixed.

PR #708, opened 2023-12-04
IGVF-1268-originated-from-desc

ian-whaling:


khzlin (changes requested):
we don't have "dedifferentiation" classification

khzlin (approved):

ian-whaling:
> we don't have "dedifferentiation" classification

We do not have dedifferentiation classification but a sample can be originated from another through dedifferentiation. I don't think the classification is relevant here and the description for cell_fate_change_treatments references dedifferentation the same way.

gabdank (approved):

PR #715, opened 2023-12-06
IGVF-1317-loci-regex-fix

ian-whaling:


src/igvfd/schemas/changelogs/prediction_set.md
```diff
@@ -3,6 +3,7 @@
 ### Schema version 3
 
 * Disallow empty strings in `description`.
+* Allow underscores in the pattern for `chromosome` in `loci`.
```
jenjou, line 6: Minor comment, looks like the property in prediction set is specifically "targeted_loci"

ian-whaling: fixed

jenjou (approved):

PR #727, opened 2023-12-11
IGVF-1344-embed-fileset-construct

forresttanaka:
I need this for [IGVF-1079](https://igvf.atlassian.net/browse/IGVF-1079) so that we can more efficiently retrieve the construct libraries associated with the sample objects associated with MeasurementSet or PredictionSet.

[IGVF-1079]: https://igvf.atlassian.net/browse/IGVF-1079?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/igvfd/mappings/measurement_set.json
+++ b/src/igvfd/mappings/measurement_set.json
@@ -1,6 +1,6 @@
 {
-    "hash": "4ff07a9192f27e9172a9669ae08d8b86",
-    "index_name": "measurement_set_4ff07a91",
+    "hash": "8658f881f2cf619ac12e6bed34adb152",
+    "index_name": "measurement_set_8658f881",
     "item_type": "measurement_set",
     "mapping": {
         "dynamic_templates": [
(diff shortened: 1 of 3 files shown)


forresttanaka:
I don’t understand why I needed to add to the `FileSet` `samples.sample_terms` embedding to make this embedding work. I feel like `construct_library_set` has nothing to do with `sample_terms`, but without this addition, my embedding change has no effect.

ian-whaling:
> I don’t understand why I needed to add to the `FileSet` `samples.sample_terms` embedding to make this embedding work. I feel like `construct_library_set` has nothing to do with `sample_terms`, but without this addition, my embedding change has no effect.

The reason it's included in samples.sample_terms instead of just samples is that all of the properties specified in the "include" array are embedded from both objects as you go down the path samples->sample_terms. So instead of having two lines where samples has a list of properties to embed and samples.sample_terms has one as well we can just have one line. It also avoids traversing samples twice were we to have multiple lines.

gabdank (changes requested):
i think this can go in, after mapping regeneration

ian-whaling (approved):

PR #729, opened 2023-12-12
IGVF-1337-rm-oligo-ab-file-set-type-aux-set

meenakshikagda:


src/igvfd/upgrade/file_set.py
```diff
@@ -157,3 +157,15 @@ def file_set_7_8(value, system):
...
+
+
+@upgrade_step('auxiliary_set', '5', '6')
+def auxiliary_set_5_6(value, system):
+    # https://igvf.atlassian.net/browse/IGVF-1337
+    if 'file_set_type' in value:
+        if value['file_set_type'] == 'oligo-conjugated antibodies':
+            value['file_set_type'] = 'cell hashing'
+            notes = value.get('notes', '')
+            notes += f'Original file_set_type was oligo-conjugated antibodies. This was replaced to be cell hashing during upgrade.'
```
jenjou, line 169: Minor comment, but this string should start with a space (like in line 143 above). It is just so that when you append it to the existing notes there will be a space after the last period and we don't get something like "word.Original file set..." (no space)

jenjou (dismissed):

jenjou (approved):

meenakshikagda:
A new branch created.

PR #730, opened 2023-12-13
IGVF-1346-calc-analysis-set-summary

iyoungworth:


gabdank (dismissed):

mingjiecn (changes requested):
I start to regret that I approved the previous ticket from Ian using recursion. Recursion generally has higher time complexity compared to iteration and can consume more memory than iteration. For Ian's ticket, the function is more like a binary search so it is still easy to understand and using recursion seems to be elegant. But here there is a lot of if elif and else, it is really difficult to understand. I feel that the recursion loses its elegance here. I suggest to convert recursion to iteration.

src/igvfd/types/file_set.py
```diff
@@ -125,12 +125,53 @@ class AnalysisSet(FileSet):
...
+                        if 'preferred_assay_title' in fileset_object:
+                            assay_terms.add(fileset_object['preferred_assay_title'])
+                        else:
+                            assay_terms.add(request.embed(fileset_object['assay_term'],
+                                            '@@object?skip_calculated=true')['term_name'])
+                    elif not input_fileset.startswith('/analysis-sets/'):
+                        fileset_types.add(fileset_object['file_set_type'])
+                    elif (input_fileset.startswith('/analysis-sets/') and
+                          fileset_object.get('input_file_sets', False)):
+                        filesets_to_inspect.update(set(fileset_object.get('input_file_sets')))
```
mingjiecn, line 164: for `input_file_set` in `input_file_sets`, make sure it is not in `inspected_filesets` first, than you can add it to `filesets_to_inspect`

mingjiecn (changes requested):

src/igvfd/types/file_set.py
gabdank: fixed

src/igvfd/tests/test_types_analysis_set.py
```diff
@@ -18,3 +18,82 @@ def test_calculated_donors(testapp, analysis_set_base, primary_cell, human_donor
...
+        {
+            'input_file_sets': [measurement_set_mpra['@id'],
+                                measurement_set_multiome['@id'],
+                                primary_analysis_set['@id']]
+        }
+    )
+    res = testapp.get(analysis_set_base['@id']).json
+    assert res.get('summary', '') == 'intermediate analysis of ATAC-seq, STARR-seq, lentiMPRA data'
+    # An infinite loop should be avoided
+    testapp.patch_json(
```
mingjiecn, line 92: This patch causes error: r`equest = <Request at 0x7f9634ecda50 PATCH http://localhost/analysis-sets/IGVFDS0682MRHA/> exc = RecursionError('maximum recursion depth exceeded')`. What kind of file set are you trying to generate here?

mingjiecn (changes requested):

src/igvfd/tests/test_types_analysis_set.py
gabdank: I know that the patch is causing it. that is why I am saying it is not the summary calculation to blame. I am trying to create circular relationship .
Analysis A input_file_sets = [B]
Analysis B input_file_Sets = [A]

PR #735, opened 2023-12-14
IGVF-1345-file-set-type-facet

ian-whaling:


jenjou:
Ticket looks ok to me. I am unsure how useful the facet is for Measurement Set right now, since there is only one enum, but I also understand that it would be included for consistency across the File Sets. So that seems OK to me.

jenjou (approved):

PR #736, opened 2023-12-18
IGVF-1340-certificate-schema

jenjou:


gabdank (commented):
the code looks fine, the missing part is a link to google drive of cloud bucket where the actual IC will be placed, even though it is not going to be publicly shared. the property should be import_items

src/igvfd/schemas/institutional_certificate.json
```diff
@@ -0,0 +1,117 @@
...
+            "$ref": "mixins.json#/standard_status"
+        }
+    ],
+    "dependentSchemas": {},
+    "properties": {
+        "schema_version": {
+            "default": "1"
+        },
+        "certificate_identifier": {
+            "title": "Certificate Identifier",
```
jenjou, line 36: Add import_items.

src/igvfd/schemas/institutional_certificate.json
```diff
@@ -0,0 +1,138 @@
...
+            "$ref": "mixins.json#/basic_item"
+        },
+        {
+            "$ref": "mixins.json#/attribution"
+        },
+        {
+            "$ref": "mixins.json#/standard_status"
+        }
+    ],
+    "dependentSchemas": {},
```
gabdank, line 31: i think we need a dependency to require "data_use_limitation" if the controlled access=true

gabdank (changes requested):

gabdank (approved):

gabdank (changes requested):
i think you need to add search configs

gabdank (approved):

PR #740, opened 2024-01-08
IGVF-1369-seq-barcodes

jenjou:


gabdank (changes requested):
maybe the test can make sets and compare them, rather than try to compare items in lists?

gabdank (approved):

PR #749, opened 2024-01-17
IGVF-1339-nic-audit

jenjou:


src/igvfd/audit/measurement_set.py
```diff
@@ -182,3 +182,57 @@ def audit_preferred_assay_title(value, system):
...
+    award = value.get('award')
+    samples = value.get('samples')
+
+    for s in samples:
+        sample_object = system.get('request').embed(s, '@@object')
+        nic_labs = []
+        nic_awards = []
+        for nic in sample_object.get('institutional_certificates', []):
+            nic_object = system.get('request').embed(nic, '@@object?skip_calculated=true')
+
```
gabdank, line 228: i think this is unnecessary empty line 228?

gabdank (changes requested):
remove extra empty line

gabdank (approved):

src/igvfd/audit/measurement_set.py
```diff
@@ -182,3 +182,47 @@ def audit_preferred_assay_title(value, system):
...
+@audit_checker('MeasurementSet', frame='object')
+def audit_inconsistent_institutional_certification(value, system):
+    '''
+        audit_detail: Measurement sets for mapping assays are expected to link to samples covered by an institutional certificate issued to a matching lab and award.
+        audit_category: inconsistent institutional certification
+        audit_levels: ERROR
+    '''
+    # Characterization assays do not need to be audited.
+    characterization_assays = [
```
gabdank, line 195: if we are going for assay_temr objects, I think it would be more future proof to switch to IDs rather than terms, even though the code will be less "readable" the term could change in ontology, the ID should be more stable

gabdank (changes requested):
suggesting to use the term_id instead of term

src/igvfd/audit/measurement_set.py
```diff
@@ -182,3 +182,47 @@ def audit_preferred_assay_title(value, system):
...
+    ]
+    assay_term = value.get('assay_term')
+    assay_object = system.get('request').embed(assay_term, '@@object?skip_calculated=true')
+    assay_term_id = assay_object.get('term_id')
+    if assay_term_id in characterization_assays:
+        return
+
+    lab = value.get('lab')
+    award = value.get('award')
+    samples = value.get('samples')
```
gabdank, line 212: should not for each of these gets be a default value returned? 
samples = value.get('samples', [])
  lab = value.get('lab', None)
    award = value.get('award', None)

gabdank (changes requested):

gabdank (approved):

PR #752, opened 2024-01-23
IGVF-1390-release-timestamp

ian-whaling:


src/igvfd/types/base.py
```diff
@@ -276,8 +276,8 @@ def _update_status(self, new_status, current_status, current_properties, schema,
         # Some release specific functionality.
         if new_status == 'released':
             # This won't be reassigned if you rerelease something.
-            if 'date_released' in schema['properties'] and 'date_released' not in new_properties:
-                new_properties['date_released'] = str(datetime.now().date())
+            if 'release_timestamp' in schema['properties'] and 'release_timestamp' not in new_properties:
+                new_properties['release_timestamp'] = str(datetime.now().date())
```
gabdank, line 280: is not datetime.now().date() gives you the date only, while we want date+time?

gabdank (changes requested):
I think one thing to be fixed in base.py

gabdank (approved):

gabdank (approved):

PR #755, opened 2024-01-25
IGVF-1392-set-status-lists

jenjou:


keenangraham (dismissed):
https://github.com/IGVF-DACC/igvfd/blob/e66ab6cc99fcb06fa5d2d223d537ea989c531a1e/src/igvfd/types/base.py#L86-L113

The uploading/content error stuff should probably be removed since these aren't statuses anymore.

src/igvfd/types/donor.py
```diff
@@ -30,6 +30,9 @@ class Donor(Item):
              'term_name', 'quantity', 'quantity_units', 'observation_date'])
     ]
 
+    set_status_up = []
```
gabdank, line 33: I think we should add set_status_up documents

src/igvfd/types/file.py
```diff
@@ -103,6 +103,13 @@ class File(Item):
         Path('submitted_by', include=['@id', 'title']),
     ]
 
+    set_status_up = [
+        'derived_from',
+        'file_format_specifications',
+        'file_set'
```
gabdank, line 109: I am not sure about file_set. it is potentially possible to release a file in the archived file-set - and it should not trigger the release of the archived file set?

src/igvfd/types/modification.py
```diff
@@ -25,6 +25,9 @@ class Modification(Item):
         Path('submitted_by', include=['@id', 'title']),
     ]
 
+    set_status_up = []
+    set_status_down = []
```
gabdank, line 29: add documents to set_up

src/igvfd/types/sample.py
```diff
@@ -74,6 +74,12 @@ class Sample(Item):
         'sample_terms'
     ]
 
+    set_status_up = [
+        'construct_library_sets',
```
gabdank, line 78: add documents

gabdank: add cell fate protocol as well

gabdank (changes requested):

src/igvfd/types/sample.py
jenjou: Cell fate change protocol is included only in the additional list of In Vitro System (line 494). I didn't want to include it in the parent Sample class since the property is specific to In Vitro System but I can also move it to Sample if that is simpler. What would make more sense?

I will add Documents to the Sample list.

src/igvfd/types/sample.py
```diff
@@ -451,6 +470,8 @@ class PrimaryCell(Biosample):
     schema = load_schema('igvfd:schemas/primary_cell.json')
     embedded_with_frame = Biosample.embedded_with_frame
     audit_inherit = Biosample.audit_inherit
+    set_status_up = Biosample.set_status_up + []
```
gabdank, line 473: this is a general comment on samples - i think we should add "Treatment" to the "up" lists

gabdank (changes requested):
few comments

PR #758, opened 2024-01-26
IGVF-1394-classifications-plural

jenjou:


gabdank (changes requested):
it looks like classifications in samples that are not in vitro system have not been updated in schemas, but in search configs they were? maybe the change in schemas was missed, or I am not sure I understand the logic.

gabdank (approved):

src/igvfd/schemas/in_vitro_system.json
```diff
@@ -148,6 +148,235 @@
...
+                                            ]
+                                        }
+                                    }
+                                }
+                            }
+                        },
+                        {
+                            "properties": {
+                                "classifications": {
+                                    "contains": {
```
khzlin, line 351: remove this line?

gabdank (dismissed):

PR #772, opened 2024-02-07
IGVF-1430-CLS-auxiliary-audit

ian-whaling:


src/igvfd/audit/construct_library_set.py
```diff
@@ -75,3 +75,23 @@ def audit_construct_library_set_scope(value, system):
             )
             yield AuditFailure('inconsistent scope metadata',
                                detail, level='WARNING')
+
+
```
gabdank, line 79: I suggest to introduce a file with common functions and import the function listing all the "non sequence/configuration" files and use this function in auxiliary and library sets audits, otherwise you introduce code duplication and challenges to maintain

gabdank (changes requested):
avoid code duplication

src/igvfd/audit/file_set.py
```diff
@@ -8,6 +8,14 @@
 )
 
 
+def check_for_seq_config_files(value):
```
gabdank, line 11: can we rename the function to be `find_non_config_sequence_files(file_set)`

ian-whaling: yes, renamed

gabdank (approved):

PR #776, opened 2024-02-09
IGVF-1426-remove-genes-loci

ian-whaling:


src/igvfd/upgrade/file_set.py
```diff
@@ -202,3 +202,10 @@ def prediction_set_3_4(value, system):
 def construct_library_set_3_4(value, system):
     # https://igvf.atlassian.net/browse/IGVF-1311
     return
+
+
+@upgrade_step('construct_library_set', '4', '5')
```
gabdank, line 207: i dont see removal of genes, loci in the upgrade

gabdank (changes requested):

gabdank (dismissed):

PR #777, opened 2024-02-12
IGVF-1417-ref-file-update

jenjou:


src/igvfd/schemas/changelogs/reference_file.md
```diff
@@ -3,6 +3,7 @@
 ### Minor changes since schema version 10
 
 * Add `release_timestamp`.
```
gabdank, line 5: i think we need to indicate that some content types are restricted-deprecated?

gabdank (changes requested):
i think we should say some content types are deprecate/restricted in the changelog

gabdank (approved):

PR #784, opened 2024-02-14
IGVF-1433-expand-readout-audit

ian-whaling:


src/igvfd/audit/measurement_set.py
```diff
@@ -88,9 +88,9 @@ def audit_unspecified_protocol(value, system):
 @audit_checker('MeasurementSet', frame='object')
 def audit_inconsistent_readout(value, system):
     '''
-        audit_detail: CRISPR-based and MPRA assays are required to specify a readout, other assays should not include one.
+        audit_detail: CRISPR-based and MPRA assays are required to specify a readout, other assays should not include one. Also readout is not expected be the same as assay term.
```
gabdank, line 91: instead of: "audit_detail: CRISPR-based and MPRA assays are required to specify a readout, other assays should not include one. Also readout is not expected be the same as assay term."

 "audit_detail: CRISPR-based and MPRA assays are required to specify a readout that is different from the assay term of the measurement set, other assays should not include readout specification."

src/igvfd/tests/test_audit_measurement_set.py
```diff
@@ -196,6 +197,19 @@ def test_audit_readout(
         for error in res.json['audit'].get('NOT_COMPLIANT', [])
     )
 
+    # Assay term should not be equivalent to
```
gabdank, line 200: I think the comment is not finished? should not it include "equivalent to readout term"?

src/igvfd/audit/measurement_set.py
```diff
@@ -111,6 +111,12 @@ def audit_inconsistent_readout(value, system):
                 f'not expected to specify a data readout.'
             )
             yield AuditFailure('inconsistent readout', detail, level='NOT_COMPLIANT')
+    if assay_term == value.get('readout'):
```
gabdank, line 114: I think this condition belongs before line 107 - you are looking on an assay with readout, and it is one of the assays that are expected to have readout - so you check they are not identical.

gabdank (changes requested):
see my comments on the Pr

src/igvfd/audit/measurement_set.py
ian-whaling: Fixed.

src/igvfd/audit/measurement_set.py
ian-whaling: Fixed

src/igvfd/tests/test_audit_measurement_set.py
ian-whaling: My bad, fixed.

src/igvfd/tests/test_audit_measurement_set.py
```diff
@@ -196,6 +197,19 @@ def test_audit_readout(
         for error in res.json['audit'].get('NOT_COMPLIANT', [])
     )
 
+    # Assay term should not be equivalent to readout term
```
gabdank, line 200: i think the test should cover both conditions of the audit, assay requiring readout and assay that should not have a readout

gabdank (changes requested):

src/igvfd/tests/test_audit_measurement_set.py
ian-whaling: There are already tests above for this. I think every audit of the 3 cases is tested, not sure what else to test.

gabdank (approved):

gabdank (approved):

PR #786, opened 2024-02-16
IGVF-1462-nic-no-mouse

jenjou:


src/igvfd/audit/measurement_set.py
```diff
@@ -191,6 +191,15 @@ def audit_inconsistent_institutional_certification(value, system):
         audit_category: inconsistent institutional certification
```
gabdank, line 191: can the audit_detail say "human" somewhere, to make clear that the requirement is for human origin samples only?

gabdank (changes requested):
minor comment

gabdank (dismissed):

PR #789, opened 2024-02-17
IGVF-1467-submitted-files-timestamp

gabdank:


src/igvfd/types/file_set.py
```diff
@@ -41,7 +43,8 @@ class FileSet(Item):
         Path('lab', include=['@id', 'title']),
         Path('submitted_by', include=['@id', 'title']),
         Path('files', include=['@id', 'accession', 'aliases', 'content_type',
-             'file_format', 'file_size', 'href', 's3_uri', 'submitted_file_name']),
+             'file_format', 'file_size', 'href', 's3_uri', 'submitted_file_name',
+                               'created_timestamp']),
```
jenjou, line 47: I think this may be a typo, "creation_timestamp". It looks like the property name is correct below in line 131

jenjou (changes requested):
I think there's just one minor typo. The rest looks OK to me.

src/igvfd/types/file_set.py
gabdank: fixed

jenjou (approved):

PR #791, opened 2024-02-21
IGVF-1448-environmental-treatment

ian-whaling:


src/igvfd/schemas/treatment.json
```diff
@@ -299,7 +360,8 @@
             "type": "string",
```
gabdank, line 360: Add comment in the treatment type, providing examples of chemical, protein and environmental treeatments.

src/igvfd/schemas/treatment.json
```diff
@@ -128,34 +127,95 @@
...
+                            "required": [
+                                "treatment_term_id"
+                            ]
                         }
                     }
                 }
             ]
+        },
+        "treatment_term_name": {
+            "comment": "Environmental treatments are restricted to certain term names including: stiffness.",
```
gabdank, line 199: Maybe "restricted to term names that include currently only stiffness."

gabdank (changes requested):
add changelog update (minor changes) and other comments on PR

src/igvfd/schemas/treatment.json
ian-whaling: There already exist examples for chemical & protein, so I just added one for environmental.

src/igvfd/schemas/treatment.json
ian-whaling: done.

gabdank (approved):

gabdank (changes requested):
update changelog

gabdank (approved):

PR #808, opened 2024-02-29
IGVF-1492-restrict-fiducial-alignment

jenjou:


gabdank (changes requested):
I think it justifies schema number bump (so an upgrade), as you are preventing future submission of objects with feducials

gabdank (approved):

PR #815, opened 2024-03-05
IGVF-1513-remodel-docstrings

ian-whaling:


src/igvfd/commands/make_audit_docstring_json.py
```diff
@@ -52,53 +52,14 @@ def get_audit_function_names_from_module(module):
...
 def get_docstring_dict_from_function_name(function_name):
     docstring = eval(function_name + '.__doc__')
     if docstring is not None:
-        dict_docstring = parse_string_to_dictionary(docstring)
-        return {function_name: dict_docstring}
+        try:
+            docstring = json.loads(docstring)
+            return {function_name: docstring}
+        except:
+            raise ValueError(f'Error loading audit docstring: {docstring} for {function_name} into JSON-format.')
```
ottojolanki, line 62: I wonder if more to the point would be something like "Docstring: {docstring} in function: {function_name} is not valid JSON format}"?

ottojolanki (commented):
I think this looks good. I would maybe change the error message for invalid docstring a little bit to explicitly tell the developer what caused the error.

src/igvfd/commands/make_audit_docstring_json.py
ian-whaling: Sure I'll change it.

ottojolanki (dismissed):
LGTM

src/igvfd/audit/measurement_set.py
```diff
@@ -88,9 +96,18 @@ def audit_unspecified_protocol(value, system):
 @audit_checker('MeasurementSet', frame='object')
 def audit_inconsistent_readout(value, system):
     '''
-        audit_detail: CRISPR-based and MPRA assays are required to specify a readout that is different from the assay term of the measurement set, other assays should not include readout specification.
-        audit_category: inconsistent readout
-        audit_levels: ERROR, NOT_COMPLIANT
+    [
+        {
+            "audit_description": "CRISPR-based and MPRA assays are required to specify a readout, other assays should not include readout specification.",
```
gabdank, line 101: Would not the two entries be potentially separated on the Audits documentation UI? The second description is not mentioning assays and it maybe unclear?

ian-whaling: I don't really understand how it is unclear, if a readout is specified it should be different than the assay term regardless of if the assay expects a readout or not.

ian-whaling: what is your suggestion otherwise? To duplicate the description?

ottojolanki (dismissed):

PR #820, opened 2024-03-11
IGVF-1514-add-alias

meenakshikagda:


src/igvfd/schemas/changelogs/award.md
```diff
@@ -1,5 +1,9 @@
 ## Changelog for *`award.json`*
 
+### Minor changes since schema version 2
```
jenjou, line 3: Typo here, looks like current version is 4

jenjou (dismissed):

PR #829, opened 2024-03-22
IGVF-1533-schema-cleanup

jenjou:


gabdank (changes requested):
i think file changelogs are missing?

gabdank (approved):

PR #830, opened 2024-03-25
IGVF-1435 add ENCODE to dbxrefs of samples

gabdank:


src/igvfd/tests/data/inserts/primary_cell.json
```diff
@@ -23,7 +23,8 @@
         "starting_amount": 5,
         "starting_amount_units": "μg",
         "dbxrefs": [
-            "GEO:SAMN45"
+            "GEO:SAMN45",
+            "ENCODE:ENCDO123AAD"
```
jenjou, line 27: The regex pattern is ENCBS, this insert has ENCDO -- probably going to cause a test failure?

gabdank: yeah, sorry let me fix

jenjou (approved):

PR #854, opened 2024-04-10
IGVF-1453-audit-status-relationship

ian-whaling:


src/igvfd/audit/item.py
```diff
@@ -53,3 +57,58 @@ def audit_item_schema(value, system):
...
+@audit_checker('Item', frame='object')
+def audit_item_status(value, system):
+    if 'status' not in value:
+        return
+
+    # objects with these statuses are not audited
+    if value['status'] in ['revoked', 'disabled', 'deleted', 'replaced']:
+        return
+
+    if value['status'] in ['archived']:
```
gabdank, line 71: is not level defined below?

gabdank (changes requested):
see the comment about definition of level

src/igvfd/audit/item.py
ian-whaling: you're right, fixed. surprised this didn't error out

gabdank (approved):

PR #860, opened 2024-04-13
IGVF-1567-reverse-link-analysis

meenakshikagda:


src/igvfd/types/file_set.py
```diff
@@ -139,6 +140,22 @@ def submitted_files_timestamp(self, request, files):
             res = sorted(timestamps, key=lambda x: datetime.strptime(x, '%Y-%m-%dT%H:%M:%S.%f%z'))
             return res[0]
 
+    @calculated_property(schema={
+        'title': 'Input FileSet For',
```
jenjou, line 144: I think we should write it as two words "File Set" in the titles and description.

jenjou (approved):

PR #862, opened 2024-04-15
IGVF-1586-aux-set-timestamp

jenjou:


src/igvfd/types/file_set.py
```diff
@@ -122,20 +122,30 @@ def control_for(self, request, control_for):
...
+    def submitted_files_timestamp(self, request, files, auxiliary_sets=[]):
+        timestamps = set()
         if files:
-            timestamps = set()
             for current_file_path in files:
                 file_object = request.embed(current_file_path, '@@object?skip_calculated=true')
                 timestamp = file_object.get('creation_timestamp', None)
                 if timestamp:
                     timestamps.add(timestamp)
+        if auxiliary_sets:
```
gabdank, line 139: i suggest to add files from auxiliary set and traverse the whole list of files once. 
otherwise you have code duplication lines 134-138 and 143-147

gabdank (changes requested):
see comments

src/igvfd/types/file_set.py
jenjou: Ok, will adjust.

PR #874, opened 2024-04-23
IGVF-1607-sequencing-kits

ian-whaling:


jenjou (commented):
I think this looks good, the only small thing I would add is using the new property in an insert (right now it's only on a fixture)

jenjou (approved):

PR #909, opened 2024-05-19
IGVF-1637-sample-modification-2

meenakshikagda:


jenjou (changes requested):
I think the code looks ok, however, we should add a changelog update for this. It would need to appear on the all the child types of biosample.

jenjou (approved):

PR #912, opened 2024-05-21
IGVF-1666-pooled-in-array

ian-whaling:


src/igvfd/types/sample.py
```diff
@@ -533,9 +533,15 @@ def parts(self, request, parts):
 
     @calculated_property(schema={
         'title': 'Pooled In',
-        'type': ['string', 'object'],
-        'description': 'The pooled samples in which this sample is included.',
-        'linkFrom': 'Biosample.pooled_from',
+        'type': 'array',
+        'description': 'The pooled samples in which this sample is included in.',
```
jenjou, line 537: Minor grammar nitpick - I think adding the "in" at the end is redundant. We could delete the earlier "in" (in "in which") if we add one at the end.

ian-whaling: I will just revert the addition

jenjou (approved):

PR #926, opened 2024-05-29
IGVF-1671-user-email-no-spaces

ian-whaling:


src/igvfd/upgrade/user.py
```diff
@@ -29,3 +29,9 @@ def user_3_4(value, system):
     if 'viewing_groups' in value:
         if len(value['viewing_groups']) < 1:
             del value['viewing_groups']
+
+
```
gabdank, line 33: i think the upgrade should not be empty but should be ready to handle the possibility of erroneous email

gabdank (changes requested):

src/igvfd/upgrade/user.py
gabdank: the upgrade should validate against the regex "pattern": "^[^\\s@]+@[^\\s@]+\\.[^\\s@]+$"

gabdank (changes requested):

gabdank (approved):

PR #930, opened 2024-05-29
IGVF-1674-refer-to-concrete

ian-whaling:


src/igvfd/audit/sample.py
```diff
@@ -25,6 +26,7 @@ def audit_sample_sorted_from_parent_child_check(value, system):
         }
     ]
     '''
+    object_type = space_in_words(system.get('types')[0]).capitalize()
```
ian-whaling, line 29: This one gets the type differently from the rest since the frame is frame='object?skip_calculated=true'

jenjou:
Looks like it needs a rebase, but code looks OK.

jenjou (dismissed):

PR #936, opened 2024-06-05
IGVF-1694-rename-fold-over

ian-whaling:


src/igvfd/schemas/changelogs/signal_file.md
```diff
@@ -1,5 +1,9 @@
 ## Changelog for *`signal_file.json`*
 
+### Schema version 9
+
+* Rename `content_type` `fold over change control` to `fold change over control`.
```
jenjou, line 5: Minor note, the changelog sentence structure should be "Adjust [property name] enum list to replace [old enum] with [new enum]." Same for the others as well.

jenjou (changes requested):
Looks ok, just requesting a small change on the changelogs.

jenjou (approved):

PR #940, opened 2024-06-08
IGVF-1701 principal analysis

gabdank:


src/igvfd/schemas/changelogs/analysis_set.md
```diff
@@ -1,5 +1,9 @@
 ## Changelog for *`analysis_set.json`*
 
+### Schema version 7
+
+* Rename `file_set_type` `primary analysis` to `principal analysis`.
```
jenjou, line 5: The changelog should follow the pattern "Adjust [property name] enum list to replace [old enum] with [new enum]."

src/igvfd/upgrade/file_set.py
```diff
@@ -330,3 +330,13 @@ def measurement_set_16_17(value, system):
...
+
+
+@upgrade_step('analysis_set', '6', '7')
+def analysis_set_6_7(value, system):
+    # https://igvf.atlassian.net/browse/IGVF-1701
+    if value.get('file_set_type', '') == 'primary analysis':
+        value['file_set_type'] = 'principal analysis'
+        notes = value.get('notes', '')
+        notes += f'This object\'s file_set_type was primary analysis and has been updated to be principal analysis.'
+        value['notes'] = notes.strip()
```
jenjou, line 342: I think it requires additional check in case "notes" is an empty string

src/igvfd/audit/analysis_set.py
```diff
@@ -14,19 +14,19 @@ def audit_input_file_sets(value, system):
     '''
     [
         {
-            "audit_description": "Primary analysis sets are expected to have at least one measurement set as an input file set.",
```
jenjou, line 17: I think this audit is being removed entirely by IGVF-1675, so perhaps this ticket should wait a little bit and rebase?

src/igvfd/upgrade/file_set.py
gabdank: no, because it is in if close where the notes will be made non empty

jenjou: I see, Ok.

jenjou (dismissed):
I think this code is OK but there will be a merge conflict in the audit file, since the entire block of that audit is getting deleted by another ticket.

PR #958, opened 2024-06-19
IGVF-1728-ont-term-summary

jenjou:


gabdank (approved):

gabdank (changes requested):
not sure if changelog should be updated as well, and also probably useful to add a test for the calculated property

PR #983, opened 2024-07-05
IGVF-1736-abstract-modification

jenjou:


keenangraham (changes requested):
Looks okay except I think the original Modification mapping needs to be manually deleted (https://github.com/IGVF-DACC/igvfd/blob/cee476760b3647a9cf2164917f92c352bead471b/src/igvfd/mappings/modification.json).

jenjou:
Deleted modification mapping. On a side note, we have mapping of sequence_data and human_genomic_variant that did not get deleted, that should probably be done (in a different ticket?)

keenangraham:
Yes, please delete here if they no longer exist.

keenangraham (dismissed):

keenangraham (approved):

PR #986, opened 2024-07-08
IGVF-1766-add-enum-prediction-set

shengchd:


src/igvfd/tests/data/inserts/prediction_set.json
```diff
@@ -9,7 +9,7 @@
         "donors": [
             "igvf:alias_human_donor1"
         ],
-        "file_set_type": "functional effect",
+        "file_set_type": "binding effect",
```
khzlin, line 12: can you create a new object in the data insert instead of replacing the `file_set_type` in an existing one.

khzlin (changes requested):

src/igvfd/tests/data/inserts/prediction_set.json
shengchd: yes will do

khzlin (approved):
looks good to me!

PR #990, opened 2024-07-10
IGVF-1787-audit-no-step-version

ian-whaling:


src/igvfd/tests/fixtures/schemas/analysis_set.py
```diff
@@ -34,17 +34,15 @@ def analysis_set_with_sample(
 
 
 @pytest.fixture
-def analysis_set_with_donor(
+def analysis_set_with_no_samples(
```
ian-whaling, line 37: This fixture was non-functional as donors are not submittable for analysis sets so I just renamed and repurposed it.

shengchd (approved):
Looks good to me!

PR #993, opened 2024-07-15
IGVF-1758-more-file-set-facets

ian-whaling:


src/igvfd/types/file_set.py
```diff
@@ -45,7 +45,7 @@ class FileSet(Item):
         Path('submitted_by', include=['@id', 'title']),
         Path('files', include=['@id', 'accession', 'aliases', 'content_type',
              'file_format', 'file_size', 'href', 's3_uri', 'submitted_file_name',
-                               'creation_timestamp', 'sequencing_platform']),
+                               'creation_timestamp', 'sequencing_platform', 'upload_status']),
         Path('control_for', include=['@id', 'accession', 'aliases']),
         Path('donors', include=['@id', 'accession', 'aliases', 'sex', 'status', 'taxa']),
         Path('samples.sample_terms', include=[
```
jenjou, line 51: I'm kind of confused about this entire block. The path is "samples.sample_terms" but it seems like all these properties are on the Sample level? Is the path supposed to be just "samples"? Or maybe I'm misunderstanding here.

ian-whaling: When embedding samples.sample_terms it will follow the path from sample_terms back to samples and include every property listed in the `include` array for both linked samples and sample_terms. So writing it this way makes it so it is only iterating through the samples once instead of twice were I to have separate lines for samples and samples.sample_terms embedding.

jenjou (dismissed):

PR #1007, opened 2024-07-23
IGVF-1829-crispr-treatment-upgrade-fix

jenjou:


src/igvfd/upgrade/crispr_modification.py
```diff
@@ -37,4 +37,5 @@ def crispr_modification_2_3(value, system):
         if 'lot_id' in value:
             notes += f' Lot_id {value["lot_id"]} was removed from this modification.'
             del value['lot_id']
-    value['notes'] = notes.strip()
+    if notes:
```
ian-whaling, line 40: Should not this if statement be added to all of the upgrades included in the previous ticket?

ian-whaling: Nvm "the other cases are “safe” because the line where I assign “notes” is within an if statement already."

ian-whaling (approved):

PR #1012, opened 2024-07-25
IGVF-1731-marker-gene-orf-feat-summ

jenjou:


ian-whaling (changes requested):
I think it looks fine but can you add tests for ORFs and phenotypic features

ian-whaling (approved):

PR #1020, opened 2024-07-29
IGVF-1795-add-enums

meenakshikagda:


src/igvfd/schemas/changelogs/reference_file.md
```diff
@@ -2,6 +2,7 @@
 
 ### Minor changes since schema version 14
 
+* Extend `content_type` enum list to include `coding_variants`, `documentation (readme)`, `variants_coding_variants`, `variants_diseases`, `variants_diseases_genes`, `variants_phenotypes` and `variants_phenotypes_studies`.
```
jenjou, line 5: We are only including one enum per line now, so this should be divided into multiple lines

jenjou (approved):

PR #1028, opened 2024-08-01
IGVF-1782-base-modifications

jenjou:


src/igvfd/schemas/sequence_file.json
```diff
@@ -296,6 +296,28 @@
                 "appscript": "ACTGGTCC",
                 "igvf_utils": "ACTGGTCC"
             }
+        },
+        "base_modifications": {
```
ian-whaling, line 300: Should this property be restricted to nanopore reads?

jenjou: That would probably make sense, I'll add that

ian-whaling (approved):

PR #1031, opened 2024-08-02
IGVF-1776-rm-library-construction

meenakshikagda:


jenjou (approved):

ian-whaling (changes requested):
Can you also remove it from facets?

ian-whaling (changes requested):
Also remove from facet_groups and search columns below.

jenjou:
Thanks @ian-whaling for catching that, I think there is one more line that needs to be removed from that file (it is in the "Facet groups" section lower in the file)

ian-whaling (dismissed):

meenakshikagda:
> Thanks @ian-whaling for catching that, I think there is one more line that needs to be removed from that file (it is in the "Facet groups" section lower in the file)

I think everything is addressed now. Am I still missing something @jenjou ?

jenjou:
> > Thanks @ian-whaling for catching that, I think there is one more line that needs to be removed from that file (it is in the "Facet groups" section lower in the file)
> 
> I think everything is addressed now. Am I still missing something @jenjou ?

No, I think it is fine. I was looking at an old version of your PR and did not refresh it so I didn't see the new commit you made

PR #1034, opened 2024-08-06
IGVF-1765-more-file-facets

ian-whaling:


src/igvfd/searches/configs/construct_library_set.py
```diff
@@ -57,6 +90,21 @@ def construct_library_set():
                     'file_set_type',
                     'selection_criteria',
                     'scope',
+                    'associated_phenotypes.term_name',
+                    'tiling_modality',
+                    'guide_type',
+                    'targeted_genes',
```
jenjou, line 96: The path here is mismatched from the above defined facet (line 25)

src/igvfd/types/file_set.py
```diff
@@ -666,16 +666,20 @@ class ConstructLibrarySet(FileSet):
         Path('lab', include=['@id', 'title']),
         Path('submitted_by', include=['@id', 'title']),
         Path('files', include=['@id', 'accession', 'aliases', 'content_type', 'file_format', 'upload_status']),
+        Path('integrated_content_files', include=['@id', 'accession',
+             'aliases', 'content_type', 'file_format', 'upload_status']),
         Path('control_for', include=['@id', 'accession', 'aliases']),
         Path('associated_phenotypes', include=['@id', 'term_id', 'term_name']),
         Path('small_scale_gene_list', include=['@id', 'geneid', 'symbol', 'name', 'synonyms']),
         Path('applied_to_samples', include=['@id', '@type', 'accession',
```
jenjou, line 674: Does this array need to include "donors" if the facet path is "applied_to_samples.donors.taxa"?

src/igvfd/searches/configs/construct_library_set.py
ian-whaling: You are correct, also it is not the right property - should have been small_scale_gene_list instead. Thanks for catching that.

src/igvfd/types/file_set.py
ian-whaling: You are correct for both, thanks for catching this. Fixed.

src/igvfd/searches/configs/construct_library_set.py
```diff
@@ -57,6 +90,21 @@ def construct_library_set():
                     'file_set_type',
                     'selection_criteria',
                     'scope',
+                    'associated_phenotypes.term_name',
+                    'tiling_modality',
+                    'guide_type',
+                    'small_scale_gene_list.symbol',
+                    'orf_list.gene',
```
forresttanaka, line 97: No facet appears for `orf_list.gene`. Should it instead show `off_list.gene.symbol`?

ian-whaling: You are correct, fixed.

src/igvfd/searches/configs/construct_library_set.py
```diff
@@ -22,6 +22,36 @@ def construct_library_set():
             'selection_criteria': {
                 'title': 'Selection Criteria'
             },
+            'small_scale_gene_list.symbol': {
+                'title': 'Targeted Genes'
```
jenjou, line 26: I think in terms of code this is fine, but it is throwing me off slightly that this facet is called Targeted Genes, meanwhile in Measurement Set we have the property "targeted_genes" with no relation to this. But I'm not sure I have a better replacement facet title either.

jenjou: I see. I think it is ok to leave it like this, just threw me off a bit. But we can always reevaluate later if there are more people getting confused.

jenjou (dismissed):

src/igvfd/searches/configs/construct_library_set.py
ian-whaling: I think we should firmly distinguish the difference between assessed_genes, targeted_genes, and small_scale_gene_list/large_scale_gene_list. But I think this can accomplished in a follow up ticket which will also update the description. Between its usage on prediction set and CLS it seems small_scale_gene_list/large_scale_gene_list is more of an "input" and the former two are "outputs". Hope that is a start.

PR #1038, opened 2024-08-07
IGVF-1754-summary-image-page

ian-whaling:


src/igvfd/schemas/changelogs/image.md
```diff
@@ -1,5 +1,9 @@
 ## Changelog for *`image.json`*
 
+### Minor changes since schema version 4
+
+* Add calculated `summary`.
```
jenjou, line 5: I think `summary` was previously already in this object (just defaulted to UUID). In other summary tickets I wrote: Update calculation of summary.

src/igvfd/types/image.py
```diff
@@ -60,3 +60,15 @@ def thumb_nail(self, request, attachment):
...
+        condition='caption',
+        schema={
+            'title': 'Summary',
+            'type': 'string',
+            'notSubmittable': True,
+        }
+    )
+    def summary(self, caption):
+        if caption:
+            return caption
```
jenjou, line 74: In both cases, I would add an else statement to return UUID if there is no caption or title. I think you may also need to write `caption=None` or `title=None` in the function args in case the property is not present since it's not required.

jenjou (changes requested):
See comments above

src/igvfd/schemas/changelogs/image.md
ian-whaling: fixed.

src/igvfd/types/image.py
ian-whaling: fixed.

jenjou: Could you also add to it to return `uuid` if no caption?

And sorry, I made a mistake about Page (will comment below)

src/igvfd/types/page.py
```diff
@@ -52,6 +52,17 @@ def unique_keys(self, properties):
...
         return keys
 
+    @calculated_property(
+        schema={
+            'title': 'Summary',
+            'type': 'string',
+            'notSubmittable': True,
+        }
+    )
+    def summary(self, title=None):
```
jenjou, line 62: I was wrong, title is required for Page. So title=None probably isn't needed here.

ian-whaling: fixed

jenjou (approved):

jenjou (changes requested):
Test is failing for some reason, not sure why yet.

jenjou (approved):

PR #1041, opened 2024-08-12
IGVF-1798-add-external-image

meenakshikagda:


src/igvfd/schemas/changelogs/analysis_set.md
```diff
@@ -1,5 +1,9 @@
 ## Changelog for *`analysis_set.json`*
 
+### Minor changes since schema version 8
+
+* Add external_image_data_url
```
jenjou, line 5: The new property should be enclosed by the `` characters, also with a period at the end. Add `external_image_data_url`.

jenjou (changes requested):
See comment, requesting update to changelog formatting.

jenjou (dismissed):

PR #1053, opened 2024-08-21
IGVF-1870-cell-type-annotation

ian-whaling:


hitz (changes requested):
I think this needs to be added to Genome Browser Annotation Files and Tabular Files (bed).  It may one day be needed for Model files but not 100% sure on that.

src/igvfd/schemas/mixins.json
```diff
@@ -623,5 +623,17 @@
                 "igvf_utils": "john-doe:measurement_set01, IGVFDS1222ABCD"
             }
         }
+    },
+    "cell_type_annotation": {
+        "cell_type_annotation": {
+            "title": "Cell Type Annotation",
+            "description": "The inferred cell type this file is associated with based on expression profiling.",
```
hitz, line 630: say "single-cell" profiling because I believe you can do this with pure snATAC data as well (although it's not trivial)

hitz (changes requested):

src/igvfd/schemas/mixins.json
ian-whaling: fixed

hitz (approved):

PR #1057, opened 2024-08-23
IGVF-1889-fix-CRISPR-summary

ian-whaling:


jenjou (changes requested):
It looks like the "test summary" for CLS is failing due to the change with the phenotypic terms. Mapping also needs update.

jenjou (approved):

PR #1058, opened 2024-08-27
IGVF-1832-control-type

meenakshikagda:


src/igvfd/schemas/measurement_set.json
```diff
@@ -163,6 +163,21 @@
                 "igvf_utils": "john-doe:measurement_set01, TSTDS98084730"
             }
         },
+        "control_type": {
+            "title": "Control Type",
+            "description": "This property describes the type of control this file-set represents.",
+            "comment": "This property is overwritten by the subclasses to define specific enum values.",
```
jenjou, line 169: Since Measurement Set is the subclass, I don't think it is necessary to include the comment again

src/igvfd/schemas/file_set.json
```diff
@@ -72,6 +72,17 @@
                 "igvf_utils": "GEO:GSE5169185, GEO:GSE5169186, GEO:GSE5169187"
             }
         },
+        "control_type": {
+            "title": "Control Type",
+            "description": "This property describes the type of control this file-set represents.",
```
jenjou, line 77: Minor suggestions: I think this can be shortened by dropping the "This property describes". And we have not used hyphen in between "file set" previously.

jenjou (changes requested):
The changelogs need to be updated for all the file set subclasses. Although it can only be specified on Measurement Set right now, I think we should still update changelog for everything.

src/igvfd/schemas/changelogs/prediction_set.md
```diff
@@ -2,6 +2,10 @@
 
 ### Minor changes since schema version 8
 
+Add `control_type`.
+
+### Minor changes since schema version 8
```
jenjou, line 7: Extra header is not needed here

src/igvfd/schemas/changelogs/analysis_set.md
```diff
@@ -2,6 +2,10 @@
 
 ### Minor changes since schema version 8
 
+Add `control_type`.
+
+### Minor changes since schema version 8
```
jenjou, line 7: Extra header is not needed here

jenjou (approved):

jenjou (approved):

PR #1063, opened 2024-08-29
IGVF-1855-add-CRISPR-assays

ian-whaling:


src/igvfd/audit/measurement_set.py
```diff
@@ -190,10 +190,9 @@ def audit_missing_institutional_certification(value, system):
     # Characterization assays do not need to be audited if they do not have controlled access data.
     characterization_assays = [
         'OBI:0003133',  # cas mediated mutagenesis
-        'NTR:0000520',  # CRISPR screen
-        'NTR:0000657',  # proliferation CRISPR screen
-        'NTR:0000722',  # CRISPR perturbation screen followed by flow cytometry and FISH
-        'NTR:0000776',  # CRISPR perturbation screen followed by single-cell RNA sequencing
+        'NTR:0000657',  # in vitro CRISPR screen assay
```
hitz, line 193: should mention to Jennifer that this is going in now, because she was suggesting we update OBI and convert these to real OBI ids.   I think I told her there was no hurry, but that could be interpreted in multiple ways.

hitz (dismissed):

src/igvfd/audit/measurement_set.py
ian-whaling: I can update them here.

PR #1065, opened 2024-09-03
IGVF-1890-model-file-set-summary

jenjou:


src/igvfd/tests/test_types_model_set.py
```diff
@@ -0,0 +1,6 @@
+import pytest
+
+
+def test_summary(testapp, model_set_no_input):
+    res = testapp.get(model_set_no_input['@id'])
+    assert res.json.get('summary') == 'predictive model v0.0.1 neural network targeting genes'
```
shengchd, line 6: would it make more sense to change the word 'targeting' to 'predicting'?

shengchd (dismissed):
code looks good, I just have a suggestion on wording.

shengchd (dismissed):

shengchd (approved):

PR #1068, opened 2024-09-04
IGVF-1840-require-pfat

ian-whaling:


jenjou (approved):
Is there a WRN ticket to get everything patched beforehand? It can be handled following upgrade as well but I think it will be a little cleaner if it can be done before.

PR #1069, opened 2024-09-04
IGVF-1885-inherit-protocols-from-measurement-set-on-analysis-set

zhwshen:


src/igvfd/schemas/analysis_set.json
```diff
@@ -92,6 +92,22 @@
                 "appscript": "https://cellpainting-gallery.s3.amazonaws.com/index.html#cpg0011-lipocyteprofiler/broad/images/Batch5/images/BR00101116/",
                 "igvf_utils": "https://cellpainting-gallery.s3.amazonaws.com/index.html#cpg0011-lipocyteprofiler/broad/images/Batch5/images/BR00101116/"
             }
+        },
+        "protocols": {
```
jenjou, line 96: When adding a calculated property, we don't want to add a non-calculated (i.e. submittable) property to the schema. This is actually clashing since now there are 2 properties with the same name. So the only change that needs to happen here is in the /types/file_set.py file.

src/igvfd/types/file_set.py
```diff
@@ -278,6 +287,33 @@ def assay_titles(self, request, input_file_sets=None):
     def donors(self, request, samples=None):
         return get_donors_from_samples(request, samples)
 
+    @calculated_property(
+        condition='protocols',
```
jenjou, line 291: This line can be removed

src/igvfd/schemas/analysis_set.json
zhwshen: Removed the `protocols` property from `analysis_set.json` in schemas.

src/igvfd/types/file_set.py
zhwshen: Removed the `conditions=` line.

jenjou:
I see that coveralls is failing, this probably means that a test should be added into /tests/test_types_analysis_set.py . I think a test that could be referenced is "test_assay_titles" in the same file. The logic should be very similar, but instead of checking assay_titles, this test should check the contents of the new property "protocols".

zhwshen:
Some of the coveralls fail are from other codes that I didn't touch (and unrelated to the current ticket). Do we typically ignore those?

src/igvfd/types/file_set.py
```diff
@@ -278,6 +287,32 @@ def assay_titles(self, request, input_file_sets=None):
...
+            'notSubmittable': True
+        }
+    )
+    def protocols(self, request, input_file_sets=None):
+        '''Calculate an array of unique protocols for all measurement sets associated with an analysis set.'''
+        protocols = set()
+        file_set_objs = get_fileset_objs_from_input_file_sets(request=request, input_file_sets=input_file_sets)
+        for file_set_obj in file_set_objs:
+            if 'MeasurementSet' in file_set_obj.get('@type'):
+                protocol = file_set_obj.get('protocols')
```
jenjou, line 311: Small change here: since protocols is not required, it is safer to add a default return value like, `file_set_obj.get('protocols', [])`

jenjou (approved):

jenjou:
FYI, looks like it needs a rebase again, but ticket looks OK to me otherwise

jenjou (approved):

PR #1074, opened 2024-09-05
IGVF-1900-new-lab-aliases

ian-whaling:


khzlin (changes requested):
can you please add a couple aliases in data inserts?

khzlin (approved):

PR #1078, opened 2024-09-09
IGVF-1856-embed-file-sets

ian-whaling:


hitz (approved):
looks good.  Kind of obnoxious that we have to repeat all this stuff in a bunch of configs, schemas AND change logs.

PR #1080, opened 2024-09-10
IGVF-1882-sample-summary

jenjou:


src/igvfd/types/file_set.py
```diff
@@ -271,6 +271,91 @@ def assay_titles(self, request, input_file_sets=None):
...
+                sorted.add(True)
+                for file_set in sample_object['file_sets']:
+                    if file_set.startswith('/measurement-sets/'):
+                        fileset_object = request.embed(file_set, '@@object?skip_calculated=true')
+                        if 'targeted_genes' in fileset_object:
+                            for gene in fileset_object['targeted_genes']:
+                                gene_object = request.embed(gene, '@@object?skip_calculated=true')
+                                targeted_genes_for_sorting.add(gene_object['symbol'])
+
+            if 'treatments' in sample_object:
```
ian-whaling, line 315: I was thinking we could lift the purpose from the treatment into the summary.
activation -> activated with ...
agonist -> agonized with ...
stimulation -> stimulated with ...

src/igvfd/types/file_set.py
```diff
@@ -271,6 +271,91 @@ def assay_titles(self, request, input_file_sets=None):
...
+            if 'targeted_sample_term' in sample_object:
+                targeted_sample_term_object = request.embed(
+                    sample_object['targeted_sample_term'], '@@object?skip_calculated=true')
+                targeted_sample_terms.add(targeted_sample_term_object['term_name'])
+
+        sample_terms_phrase = ', '.join(sample_terms)
+        sample_classifications_phrase = ', '.join(sample_classifications)
+        targeted_sample_terms_phrase = ''
+        if targeted_sample_terms:
+            targeted_sample_terms_phrase = f'induced to {", ".join(targeted_sample_terms)}'
```
ian-whaling, line 327: wouldn't it make more sense for each sample term to be induced to its target sample term instead of the set of sample terms induced to its targeted sample terms? Wondering on the logic here

ian-whaling (approved):

src/igvfd/types/file_set.py
```diff
@@ -317,6 +317,106 @@ def protocols(self, request, input_file_sets=None):
...
+                        fileset_object = request.embed(file_set, '@@object?skip_calculated=true')
+                        if 'targeted_genes' in fileset_object:
+                            for gene in fileset_object['targeted_genes']:
+                                gene_object = request.embed(gene, '@@object?skip_calculated=true')
+                                targeted_genes_for_sorting.add(gene_object['symbol'])
+
+            if 'treatments' in sample_object:
+                for treatment in sample_object['treatments']:
+                    treatment_object = request.embed(treatment, '@@object?skip_calculated=true')
+                    treatments.add(treatment_purpose_to_adjective[treatment_object['purpose']])
```
ian-whaling, line 386: This would error out if another treatment purpose is added to the enum list and the dicitionary treatment_purpose_to_adjective is not added, use .get() instead.

ian-whaling: Also can you rename this variable to treatment_purposes

ian-whaling (changes requested):
minor comment on treatments

ian-whaling (approved):

ian-whaling (approved):

PR #1083, opened 2024-09-11
IGVF-1922-genes-loci

ian-whaling:


jenjou (changes requested):
I think this just needs corresponding update to the Analysis Step input/output_file_types.

jenjou (approved):

PR #1087, opened 2024-09-12
IGVF-1848-add-auxset-enum

zhwshen:


src/igvfd/schemas/changelogs/auxiliary_set.md
```diff
@@ -2,6 +2,7 @@
 
 ### Minor changes since schema version 11
 
+* Add `MORF barcode sequencing`.
```
jenjou, line 5: This sentence is used for adding new properties, I think the one we want here is:
> Extend `file_set_type` enum list to include `...`.

src/igvfd/schemas/auxiliary_set.json
```diff
@@ -61,6 +61,7 @@
                 "circularized RNA barcode detection",
                 "gRNA sequencing",
                 "lipid-conjugated oligo sequencing",
+                "MORF barcode sequencing",
```
jenjou, line 64: This is good, I would just add a sentence in the above "comment" (line 57) to explain the new term. https://www.addgene.org/pooled-library/human-morf-library/ and https://pubmed.ncbi.nlm.nih.gov/38697106/ are the relevant info. Just want to especially include it since the "ORF" in this is not the usual "open reading frame" so it may be more confusing than average.

I propose something like:
"MORF barcode sequencing is the sequencing of the Multiplexed Overexpression of Regulatory Factors library to produce a mapping of barcodes to an overexpressed transcription factor."

jenjou (changes requested):
See comments.

src/igvfd/schemas/changelogs/auxiliary_set.md
zhwshen: Addressed.

zhwshen:
@jenjou Not sure why this push failed me on some of the Coveralls. I checked the codes but they don't seem to involve my specific ticket (AuxSet enum adding). Some fails are HTTPs and credentials. One is about audit docstring commands for functions. The last one is about Construct Library Set.

Also, one of the coveralls says "coverage remained the same at 88.9992%" but showed up as a fail.

jenjou (approved):
It is passing tests now after rebase (but needs rebase yet again...) so will approve.

PR #1089, opened 2024-09-12
IGVF-1915-starr-seq

meenakshikagda:


src/igvfd/schemas/changelogs/assay_term.md
```diff
@@ -2,6 +2,7 @@
 
 ### Minor changes since schema version 7
 
+* Extend `preferred_assay_title` enum list to include `STARR-seq`.
```
ian-whaling, line 5: preferred_assay_titles not preferred_assay_title

src/igvfd/schemas/measurement_set.json
```diff
@@ -87,56 +87,57 @@
             "description": "The custom lab preferred label for the experiment performed in this measurement set.",
             "type": "string",
             "enum": [
```
ian-whaling, line 89: I reorganized this list to be based on assay not alphabetical ordering in https://github.com/IGVF-DACC/igvfd/pull/1063/files. I had brought it up with Ben and Jennifer, my reasoning was that alphabetical ordering here is not helpful and it would be more useful for the submitters if relevant assays were grouped together in this long list of entries. Say for example, a submitter is submitting "Growth CRISPR screen" but first see "Proliferation CRISPR screen" and chooses this term instead as they are unaware of the more precise term.

meenakshikagda: It looked very strange to me - I wasn't aware we did that deliberately.

PR #1091, opened 2024-09-16
IGVF-1930-biosample-qualifier

meenakshikagda:


src/igvfd/schemas/in_vitro_system.json
```diff
@@ -590,6 +590,19 @@
                 "appscript": "DMEM with serum",
                 "igvf_utils": "DMEM with serum"
             }
+        },
+        "biosample_qualifier": {
+            "title": "Biosample qualifier",
+            "description": "Desription of cell state",
```
jenjou, line 596: Minor typo here "description". And I wonder if we should try to write additional details to explain that it is meant to be a way to capture cell states that are more precise than what can be captured by ontologies? Or if there's another better way to describe that.

src/igvfd/schemas/in_vitro_system.json
```diff
@@ -590,6 +590,19 @@
...
+            "type": "string",
```
jenjou, line 597: Should it be an array? Wondering if it is possible to have 2 different qualifiers? Seem to recall Ben mentioning it but not sure if we decided.

src/igvfd/schemas/in_vitro_system.json
```diff
@@ -590,6 +590,19 @@
...
+            "enum": [
+                "exhausted",
+                "stimulated"
```
jenjou, line 600: Another question (maybe for lab): do we need "stimulated" here or is it something that could be captured using Treatment objects?

PR #1092, opened 2024-09-16
IGVF-1932-transcriptome-ref

meenakshikagda:


ian-whaling (changes requested):
Can you add a enum_descriptions for transcriptome reference? We have enum descriptions for content types that are submittable by all users on reference files

ian-whaling (approved):

PR #1093, opened 2024-09-16
IGVF-1930-biosample-qualifier

meenakshikagda:


src/igvfd/schemas/in_vitro_system.json
```diff
@@ -590,6 +590,19 @@
                 "appscript": "DMEM with serum",
                 "igvf_utils": "DMEM with serum"
             }
+        },
+        "biosample_qualifier": {
+            "title": "Biosample qualifier",
+            "description": "Desription of cell state",
```
jenjou, line 596: Minor typo here "description". And I wonder if we should try to write additional details to explain that it is meant to be a way to capture cell states that are more precise than what can be captured by ontologies? Or if there's another better way to describe that.

src/igvfd/schemas/in_vitro_system.json
```diff
@@ -590,6 +590,19 @@
...
+            "type": "string",
```
jenjou, line 597: Should it be an array? Wondering if it is possible to have 2 different qualifiers? Seem to recall Ben mentioning it but not sure if we decided.

src/igvfd/schemas/in_vitro_system.json
```diff
@@ -590,6 +590,19 @@
...
+            "enum": [
+                "exhausted",
+                "stimulated"
```
jenjou, line 600: Another question (maybe for lab): do we need "stimulated" here or is it something that could be captured using Treatment objects?

jenjou (changes requested):
Comments copied from previous branch.

src/igvfd/schemas/in_vitro_system.json
```diff
@@ -590,6 +590,25 @@
                 "appscript": "DMEM with serum",
                 "igvf_utils": "DMEM with serum"
             }
+        },
+        "biosample_qualifier": {
+            "title": "Biosample qualifier",
```
jenjou, line 595: Qualifier should be capitalized to match other property titles

src/igvfd/schemas/in_vitro_system.json
```diff
@@ -590,6 +590,25 @@
...
+            "description": "An array of various cell state. This property provides additional information at a finer-grained level about a cell compared to what ontologies currently capture. Example exhausted T-cells",
```
jenjou, line 596: Just some minor wording suggestions:
* various cell states (plural)
* Move "about a cell" before "at a finer-grained level"
* "For example, exhausted T-cells."

Same comment on the description within "items" on line 601

jenjou (approved):

PR #1097, opened 2024-09-19
IGVF-1931-castaneus-enums

jenjou:


shengchd (commented):
Code looks good, I'm wondering if we also want to add 'Cast - m39' to the assembly enum list of small_scale_loci_list (line 566 in mixins.json), or we don't expect the lab submitting that?

Updated: seems the lab won't be using small_scale_loci_list, we can leave it then.

shengchd (approved):

PR #1107, opened 2024-09-28
IGVF-1928-multiplexed-sample

meenakshikagda:


src/igvfd/tests/data/inserts/multiplexed_sample.json
```diff
@@ -8,6 +8,10 @@
             "igvf:multiplex_alias"
         ],
         "barcode_map": "igvf:tabular_file_04_barcode_to_sample_mapping",
+        "multiplexing_method": [
+            "genetic",
+            "barcode based"
+        ],
```
shengchd, line 14: should we use plural form in property name (multiplexing_methods), since it's an array and we allow it to have multiple items?

src/igvfd/schemas/multiplexed_sample.json
```diff
@@ -98,6 +99,22 @@
                 "igvf_utils": "TSTSM02881437, john-doe:tissue01, john-doe:primary_cell02"
             }
         },
+        "multiplexing_method": {
+            "title": "Multiplexing Method",
+            "description": "The method of multiplexing and/or demultiplexing.",
```
shengchd, line 104: can you explain a bit on what would be a 'multiplexing and demultiplexing' case?

shengchd (commented):
I have a comment on the property name and a question on the description.

shengchd (dismissed):
looks good

PR #1112, opened 2024-10-02
IGVF-1977-peak-quantification

meenakshikagda:


src/igvfd/schemas/changelogs/analysis_step.md
```diff
@@ -2,6 +2,8 @@
 
 ### Minor changes since schema version 5
 
+* Extend `input_content_types` enum list to include `peak quantifications`.
+* Extend `output_content_types` enum list to include `peak quantifications`.
```
shengchd, line 6: Need to add more content type enums here

shengchd (changes requested):
Just need to update changelog under analysis step

shengchd (approved):

PR #1115, opened 2024-10-02
IGVF-1942-prevent-uppercase-emails

keenangraham:


jenjou (changes requested):
Since there was a version bump, I think the changelog needs a new entry. Probably: "Update email regex to disallow capital letters." ?

jenjou (approved):

ottojolanki (approved):
LGTM

PR #1116, opened 2024-10-03
IGVF-1956-add-VarChAMP-collection

zhwshen:


khzlin (changes requested):
please add to data insert to some object types like one of file, file set and sample.

khzlin (approved):
Looks good to me!

PR #1125, opened 2024-10-09
IGVF-1997-MPRA-outputs

ian-whaling:


src/igvfd/schemas/tabular_file.json
```diff
@@ -135,6 +135,11 @@
                 "prime editing guide RNA sequences",
                 "protein to protein interaction score",
                 "sample sort parameters",
+                "reporter elements",
```
shengchd, line 138: need to move those enums ahead to 'sample sort parameters' follow order

shengchd (changes requested):
just need minor changes on enum order

shengchd (dismissed):

PR #1133, opened 2024-10-12
IGVF-1914-calculate-taxa

meenakshikagda:


src/igvfd/types/sample.py
```diff
@@ -843,6 +843,36 @@ def summary(self, request, multiplexed_samples=None):
     def sample_terms(self, request, multiplexed_samples):
         return collect_multiplexed_samples_prop(request, multiplexed_samples, 'sample_terms')
 
+    @calculated_property(
+        define=True,
+        schema={
+            'title': 'Taxa',
+            'type': 'string',
+            'description': 'The species of the organism.',
+            'enum': [
```
jenjou, line 852: Do we need to account for Technical Sample being in a Multiplexed Sample? Because Technical Sample has submittable taxa which includes yeast. And Taxa is also not required, so Technical Sample can have "null" taxa.

jenjou (dismissed):

jenjou (approved):

PR #1145, opened 2024-10-19
IGVF-2024-fused-domain-exp

ian-whaling:


src/igvfd/types/sample.py
```diff
@@ -491,10 +491,18 @@ def summary(self, request, sample_terms, donors, sex, age, age_units=None, embry
...
+            if modification_summaries:
+                summary_terms += f' modified with {", ".join(modification_summaries)},'
+
         # construct library set overview is appended to the end of the summary
         if (construct_library_sets and
                 biosample_type in ['primary_cell', 'in_vitro_system', 'tissue', 'whole_organism']):
-            verb = 'modified with'
+            verb = 'transfected with'
             library_types = set()
             for CLS in construct_library_sets:
```
jenjou, line 507: Slightly out of scope but since we are editing this code: can we change this variable from "CLS" to something else?  I think "CLS" is a special reserved keyword in Python (similar to "self") so we probably shouldn't use it if we can avoid it. 
I think this is also used in the Technical Sample code as well.

src/igvfd/tests/test_types_analysis_set.py
```diff
@@ -188,7 +188,16 @@ def test_analysis_set_sample_summary(testapp, principal_analysis_set, measuremen
...
-    assert res.get('sample_summary', '') == 'K562 differentiated cell specimen induced to endothelial cell of vascular tree, at 1 time point(s) post change, differentiated with treatment(s), modified with a guide library, sorted on expression of MYC'
+    assert res.get('sample_summary', '') == 'K562 differentiated cell specimen induced to endothelial cell of vascular tree, at 1 time point(s) post change, differentiated with treatment(s), modified with CRISPR modifications, transfected with a guide library, sorted on expression of MYC'
+    testapp.patch_json(
+        in_vitro_differentiated_cell['@id'],
+        {
+            'modifications': [crispr_modification['@id'], degron_modification['@id']]
+        }
+    )
+    res = testapp.get(principal_analysis_set['@id']).json
+    print(res.get('sample_summary', ''))
```
jenjou, line 199: Extra print statement

ian-whaling: fixed

src/igvfd/types/sample.py
ian-whaling: fixed

jenjou (dismissed):

PR #1146, opened 2024-10-21
IGVF-2042-add-uniform-pipeline-property

khzlin:


src/igvfd/schemas/changelogs/workflow.md
```diff
@@ -3,6 +3,7 @@
 ### Minor changes since schema version 5
 
 * Extend `collections` enum list to include `VarChAMP`.
+* Add `uniform_pipeline`.
```
meenakshikagda, line 6: This should be moved on the top

meenakshikagda (commented):
One minor comment on changelog.

src/igvfd/schemas/changelogs/workflow.md
khzlin: thanks for catching that. fixed.

meenakshikagda (dismissed):

PR #1154, opened 2024-10-26
IGVF-2040-reference-external

shengchd:


src/igvfd/upgrade/file.py
```diff
@@ -337,3 +337,14 @@ def tabular_file_12_13(value, system):
...
     if notes.strip() != '':
         value['notes'] = notes.strip()
+
+
+@upgrade_step('reference_file', '14', '15')
+def reference_file_14_15(value, system):
+    # https://igvf.atlassian.net/browse/IGVF-2040
+    if 'external_id' in value:
+        external_id = value['external_id']
+        notes = value.get('notes', '')
```
meenakshikagda, line 347: Shouldn't we initialize the notes outside of the if statement?

meenakshikagda (dismissed):

PR #1164, opened 2024-10-31
IGVF-1989-rm-model-zoo

shengchd:


src/igvfd/upgrade/file_set.py
```diff
@@ -452,3 +452,15 @@ def analysis_set_8_9(value, system):
...
+
+
+@upgrade_step('model_set', '4', '5')
+def model_set_4_5(value, system):
+    # https://igvf.atlassian.net/browse/IGVF-1989
+    notes = value.get('notes', '')
+    if 'model_zoo_location' in value:
+        model_zoo_location = value['model_zoo_location']
+        value['url'] = model_zoo_location
+        notes += f' This model set has {model_zoo_location} in model_zoo_location, but the property model_zoo_location were removed via upgrade and should be submitted in url.'
```
jenjou, line 464: Add a `.strip()` to remove extra white space (in case the object does not have any `notes` prior to this upgrade).

Also, I think the sentence can also be shortened, but depends how we decide on whether it will get added to `url` or not. Maybe something like one of these? 
* The `model_zoo_location` of this model set was moved to `url`. 
* This model set previously linked to [link] in `model_zoo_location`, which has been removed via upgrade.

PR #1165, opened 2024-11-01
IGVF-2091-new-base-mod-enums

jenjou:


shengchd (commented):
The code is good, can we also add some example data in data/inserts/?

shengchd (approved):

PR #1171, opened 2024-11-07
IGVF-2128-analysis-set-summary-fix

jenjou:


jenjou:
This is failing coveralls due to the "failsafe return value" condition, which I did not write a test for since it is supposed to be a "backup" just in case there is some condition we did not catch. But for real examples of Analysis Set, I don't think this should ever trigger.

zhwshen (approved):
LGTM

PR #1174, opened 2024-11-11
IGVF-2067-file-summary-updates

jenjou:


shengchd (changes requested):
The code is good, I just have a request to add those changes to changelog, since we are not updating all types of files here, and it would be easier to track later.

shengchd (dismissed):

PR #1177, opened 2024-11-11
IGVF-2138-fix-matrix-audit

ian-whaling:


src/igvfd/tests/test_audit_matrix_file.py
```diff
@@ -27,3 +27,14 @@ def test_audit_identical_dimensions(
         error['category'] != 'inconsistent dimensions'
         for error in res.json['audit'].get('WARNING', [])
     )
+    testapp.patch_json(
+        matrix_file['@id'],
```
zhwshen, line 31: Should this be `matrix_file` patched with both file_format and identical dimensions OR `matrix_file_hic` with a new file_format? Because the audit is to avoid flagging .cool or .mcool files with identical dimensions errors, `matrix_file` doesn't have that issue to start with. So something like 

`testapp.patch_json(matrix_file['@id'], {'file_format': 'cool', 'principal_dimension': 'gene'}` OR`testapp.patch_json(matrix_file_hic['@id'], {'file_format': 'cool'}`

ian-whaling: matrix_file had been patched prior in lines 14-19 above to have 'principal_dimension': 'gene', so it already has the audit. By patching with file_format: cool I am removing the audit since .cool files aren't flagged.

zhwshen (approved):
LGTM

PR #1182, opened 2024-11-12
IGVF-2117-gencode-32

meenakshikagda:


src/igvfd/schemas/changelogs/curated_set.md
```diff
@@ -2,6 +2,7 @@
 
 ### Minor changes since schema version 8
 
+* Extend `transcriptome_annotation` enum list to include `GENCODE 32`.
```
jenjou, line 5: I think it is a calculated property in Curated Set `transcriptome_annotations` so we may not actually need this

jenjou (dismissed):

PR #1184, opened 2024-11-12
IGVF-2137-image-status

meenakshikagda:


jenjou (approved):
I don't think we have had a situation of removing default value befo

PR #1215, opened 2024-11-21
IGVF-2179-barcode-onlist

meenakshikagda:


src/igvfd/schemas/tabular_file.json
```diff
@@ -109,6 +109,7 @@
         "content_type": {
             "comment": "Content Type describes the content of the file.",
             "enum": [
+                "barcode onlist",
```
jenjou, line 112: Can we add an entry as well below in the "enum descriptions" for the new term?

jenjou (changes requested):
Code looks fine, just requesting an enum definition to be added.

jenjou (approved):

PR #1224, opened 2024-11-25
IGVF-2188-variant-effects

meenakshikagda:


src/igvfd/schemas/measurement_set.json
```diff
@@ -128,7 +128,7 @@
                 "Perturb-seq",
                 "CERES-seq",
                 "TAP-seq",
-                "Variant FlowFISH",
+                "Variant EFFECTS",
```
ian-whaling, line 131: Shouldn't it be "Variant-EFFECTS" with a hyphen?

ian-whaling (changes requested):
Requested clarification on the name.

src/igvfd/schemas/measurement_set.json
meenakshikagda: We didn't have a hyphen previously for "Variant-FlowFISH". I can add it here though.

ian-whaling (approved):

PR #1225, opened 2024-11-25
IGVF-2182-status-embed

meenakshikagda:


jenjou (approved):
As confirmed by Forrest, status is not needed for lab, award, or submitted_by so their embedding is not altered here.

PR #1230, opened 2024-11-27
IGVF-2010-meas-set-summary-update

jenjou:


shengchd (changes requested):
Code looks good, just a request to update the changelog under measurement set.

shengchd (dismissed):

src/igvfd/types/file_set.py
```diff
@@ -825,12 +827,28 @@ def summary(self, request, assay_term, preferred_assay_title=None, samples=None)
...
             assay = f'{assay} ({preferred_assay_title})'
         else:
             assay = preferred_assay_title
 
+        # Special case for Y2H assays.
+        if request.embed(assay_term)['term_id'] == 'OBI:0000288':
+            if control_type and control_type == 'pre-selection':
+                control_phrase = control_type
+            else:
+                control_phrase = 'post-selection'
```
hitz, line 843: I understand this is just a kludge, but it seems like for Y2H there are 3 possibly things:
1) control type = pre-selection
2) control type != pre-selection but not NULL
3) control type null.
last 2 will both get "post selection" is that what is intended?  There is no Y2H data on the portal

jenjou: I think case #2 isn't supposed to happen because "pre-selection" is the only one applicable for them.
But I think it makes sense to add that case, to make it more evident when they do have something wrongly patched.

hitz (approved):

PR #1235, opened 2024-12-04
IGVF-2013-dbxrefs-isgr

meenakshikagda:


src/igvfd/schemas/human_donor.json
```diff
@@ -169,6 +169,20 @@
                 "comment": "Cell line donors should specify as so. e.g. 'K562 cell line donor'.",
                 "type": "string"
             }
+        },
+        "dbxrefs": {
+            "comment": "Supported external resources include: (The International Genome Sample Resource) ISGR.",
```
ian-whaling, line 174: I don't think this line is needed, but if you want to include you should update with the other external databases.

meenakshikagda: updated

ian-whaling (approved):

PR #1261, opened 2025-01-10
IGVF-2249-enum-sig-tab

shengchd:


khzlin (dismissed):
this looks good to me!

khzlin (approved):

PR #1263, opened 2025-01-11
IGVF-2196-add-sample-model

shengchd:


ian-whaling (changes requested):
Can you update the changelog?

src/igvfd/schemas/changelogs/model_set.md
```diff
@@ -2,6 +2,8 @@
 
 ### Minor changes since schema version 4
 
+* Add `donors`
+* Add `samples`.
```
ian-whaling, line 6: The properties were already there, just not submittable. I think something like Allow submission of `samples` or something like that would be better.

ian-whaling (changes requested):
reword changelog request

src/igvfd/schemas/changelogs/model_set.md
shengchd: yes will do, was wondering the same thing just didn't find a previous example.

ian-whaling (dismissed):

ian-whaling (approved):

PR #1266, opened 2025-01-14
IGVF-2273-embed-strain-ethnic

ian-whaling:


src/igvfd/types/sample.py
```diff
@@ -92,7 +92,8 @@ class Sample(Item):
              '@id', 'summary', 'status', 'tagged_proteins', 'modality', 'fused_domain', 'symbol', 'cas', 'cas_species', 'degron_system']),
         Path('institutional_certificates', include=['@id', 'certificate_identifier', 'status']),
         Path('construct_library_sets.associated_phenotypes', include=[
-             '@id', 'accession', 'file_set_type', 'term_name', 'associated_phenotypes', 'status'])
+             '@id', 'accession', 'file_set_type', 'term_name', 'associated_phenotypes', 'status']),
+        Path('donors', include=['@id', 'accession', 'strain', 'ethnicities']),
```
forresttanaka, line 96: Could you add an embed for the `status` property in case we display these in a table?

forresttanaka (changes requested):
We’re trying to add `status` to objects in embedded arrays since we pretty much always display that in tables.

src/igvfd/types/sample.py
ian-whaling: yes

PR #1272, opened 2025-01-16
IGVF-2261-adjust-seqspec-audit

ian-whaling:


src/igvfd/audit/file_set.py
```diff
@@ -123,7 +154,7 @@ def audit_inconsistent_seqspec(value, system):
     '''
     [
         {
-            "audit_description": "Sequence files in a file set from the same sequencing run, flowcell_id, lane, and index are expected to link to the same seqspec file, which should be unique to that set of sequence files.",
+            "audit_description": "Sequence files in a file set from the same sequencing run (with the exception of non-single cell data), flowcell id, lane, and index, the combination of which define a sequencing set, are expected to link to the same seqspec file, which should be unique to that set of sequence files.",
```
ian-whaling, line 157: I am open to suggestions on how to alternatively word this mouthful.

jenjou: I don't know if it's any clearer by moving the list of properties into its own sentence? The description becomes much longer here, though.
> Sequence files in a file set from the same  sequencing set, are expected to link to the same seqspec file, which should be unique to that set of sequence files. A sequencing set is defined by the combination of flowcell id, lane, and index for non-single cell data, or flowcell ID, lane, index, and sequencing run for single cell data.

ian-whaling: thanks

src/igvfd/audit/file_set.py
```diff
@@ -123,7 +155,7 @@ def audit_inconsistent_seqspec(value, system):
     '''
     [
         {
-            "audit_description": "Sequence files in a file set from the same sequencing run, flowcell_id, lane, and index are expected to link to the same seqspec file, which should be unique to that set of sequence files.",
+            "audit_description": "Sequence files in a file set from the same sequencing set, are expected to link to the same seqspec file, which should be unique to that set sequencing set. A sequencing set is defined by the combination of flowcell ID, lane, and index for non-single cell data, or flowcell ID, lane, index, and sequencing run for single cell data.",
```
jenjou, line 158: Could we delete that first comma after "same sequencing set"? I think I typoed in my original comment, sorry! And I think I left in an extra "sequencing set" in "that set sequencing set".

jenjou (dismissed):

PR #1274, opened 2025-01-21
IGVF-2286-add-embeddings

khzlin:


src/igvfd/tests/test_types_analysis_step.py
```diff
@@ -8,4 +8,4 @@ def test_analysis_step_name_calcprop(testapp, analysis_step):
 
 def test_analysis_step_versions(testapp, analysis_step, analysis_step_version):
     res = testapp.get(analysis_step['@id'])
-    assert res.json.get('analysis_step_versions') == [analysis_step_version['@id']]
+    assert analysis_step_version['@id'] in res
```
ian-whaling, line 11: I think the test should be something like: assert set([analysis_step_version_item['@id'] for analysis_step_version_item in res.json.get('analysis_step_versions')]) == {analysis_step_version['@id']} to be consistent with other tests

khzlin: done

src/igvfd/types/analysis_step.py
```diff
@@ -31,6 +31,8 @@ class AnalysisStep(Item):
         Path('parents', include=['@id', 'title', 'status']),
         Path('workflow', include=['@id', 'accession', 'status']),
         Path('submitted_by', include=['@id', 'title']),
+        Path('analysis_step_versions.software_versions', include=[
+             '@id', 'analsyis_step_versions', 'software_versions', 'name'])
```
jenjou, line 35: There's a typo here, although it seems like the embedding isn't broken, but should probably fix.

khzlin: thanks for catching that. fixed

ian-whaling (approved):

PR #1288, opened 2025-01-28
IGVF-2158-targeted-genes-meas-set

ian-whaling:


src/igvfd/types/file_set.py
```diff
@@ -835,6 +836,23 @@ def summary(self, request, assay_term, preferred_assay_title=None, samples=None,
         else:
             assay = preferred_assay_title
 
+        if targeted_genes:
+            # Special case for CRISR screens using flow cytometry
```
jenjou, line 840: Typo in "CRISPR"

jenjou (commented):
I think this looks OK, there's a typo in a comment but it doesn't affect the code.

src/igvfd/types/file_set.py
ian-whaling: fixed

jenjou (approved):
Looks good, although there was another merge so will require a rebase.

PR #1289, opened 2025-01-28
IGVF-2239-calc-soft-vers

ian-whaling:


src/igvfd/searches/configs/workflow.py
```diff
@@ -7,7 +7,7 @@
 def workflow():
     return {
         'facets': {
-            'analysis_steps.analysis_step_versions.software_versions.software.name': {
+            'analysis_steps.analysis_step_versions.software_versions.software.title': {
```
ian-whaling, line 10: For consistency I changed this to title as well. Don't think name makes sense vs. title

shengchd (commented):
The calculation part looks good, but I think we will also need to remove `software_version` from model set, which should be an upgrade?

src/igvfd/types/file_set.py
```diff
@@ -969,6 +970,30 @@ def externally_hosted(self, request, files=None):
...
+                    file_object = request.embed(file, '@@object?skip_calculated=true')
+                    analysis_step_version = file_object.get('analysis_step_version', '')
+                    if analysis_step_version:
+                        analysis_step_version_object = request.embed(
+                            analysis_step_version, '@@object?skip_calculated=true')
+                        software_versions = software_versions + \
+                            analysis_step_version_object.get('software_versions', [])
+        if software_versions:
+            return software_versions
+
```
shengchd, line 996: Maybe do` set()` before return and require software_versions has unique items? (e.g. bluestarr has two model files under the same model set)

ian-whaling: good idea :^)

src/igvfd/schemas/changelogs/analysis_step.md
```diff
@@ -4,6 +4,8 @@
 
 * Extend `output_content_types` enum list to include `sample barcode count matrix`.
 * Extend `input_content_types` enum list to include `sample barcode count matrix`.
+* Extend `analysis_step_types` enum list to include `computational model training`.
```
jenjou, line 7: The newer changelog entries should be at the top (probably just a mix up during rebasing?)

ian-whaling: yes, fixed

src/igvfd/schemas/changelogs/model_set.md
```diff
@@ -1,5 +1,10 @@
 ## Changelog for *`model_set.json`*
 
+### Schema version 5
+
+* Add `software_versions`.
```
shengchd, line 5: can be slightly more specific: Add calculated property ...

shengchd (changes requested):
Just a request on wording in changelog, otherwise looks good to me!

shengchd (approved):

PR #1290, opened 2025-01-28
IGVF-2276-calc-cls

ian-whaling:


src/igvfd/types/file_set.py
```diff
@@ -1196,6 +1234,32 @@ class ConstructLibrarySet(FileSet):
...
+    def file_sets(self, request, applied_to_samples=None):
+        file_sets = []
+        for sample in applied_to_samples:
+            sample_object = request.embed(sample,
+                                          '@@object_with_select_calculated_properties?'
+                                          'field=file_sets'
+                                          )
+            file_sets = file_sets + sample_object.get('file_sets', [])
+        if file_sets:
+            return file_sets
```
shengchd, line 1261: do you think set() would need to be applied to file_sets at the end? (I suppose multiple samples could be linked to same analysis set?)

src/igvfd/types/file_set.py
```diff
@@ -190,6 +191,34 @@ def submitted_files_timestamp(self, request, files, auxiliary_sets=[]):
...
+        })
+    def construct_library_sets(self, request, samples=None):
+        construct_library_sets = set()
+        for sample in samples:
+            sample_object = request.embed(sample,
+                                          '@@object_with_select_calculated_properties?'
+                                          'field=construct_library_sets'
+                                          )
+            if sample_object.get('construct_library_sets', []):
+                construct_library_sets = construct_library_sets | set(sample_object.get('construct_library_sets', []))
```
shengchd, line 218: are you only keeping the CSL for the first sample in samples?

src/igvfd/types/file_set.py
```diff
@@ -376,6 +405,16 @@ def assay_titles(self, request, input_file_sets=None):
...
+                elif 'ConstructLibrarySet' in file_set_object.get('@type'):
+                    for sample in file_set_object.get('applied_to_samples', []):
+                        sample_object = request.embed(
+                            sample, '@@object_with_select_calculated_properties?field=file_sets')
+                        for file_set in sample_object.get('file_sets', []):
+                            file_set_object = request.embed(
+                                file_set, '@@object_with_select_calculated_properties?field=preferred_assay_title')
+                            preferred_assay_title = file_set_object.get('preferred_assay_title')
+                            if preferred_assay_title:
+                                assay_titles.add(preferred_assay_title)
```
shengchd, line 417: I was thinking if we want to calculate preferred_assay_title from input_file_sets instead? seems more straightforward to me but maybe that works equivalently.

src/igvfd/types/file_set.py
ian-whaling: construct_library_sets = construct_library_sets | set(sample_object.get('construct_library_sets', [])) adds to the existing set

src/igvfd/types/file_set.py
ian-whaling: This calculated property, assay_titles, does calculate from the input_file_sets, in this case it is calculating from input construct library sets, which need to first go looking for the associated measurement sets.

src/igvfd/types/file_set.py
ian-whaling: Actually I will remove this calculated property, I realize it is unnecessary since I have embedded file sets from the samples.

src/igvfd/types/file_set.py
shengchd: oh thanks, forgot it's the union function for set

src/igvfd/types/file_set.py
shengchd: got it!

src/igvfd/schemas/changelogs/analysis_set.md
```diff
@@ -2,6 +2,7 @@
 
 ### Minor changes since schema version 9
 
+* Add calculated property `construct_library_sets`.
```
shengchd, line 5: can you also add 'update calculation of assay_titles' and summary?

shengchd: actually only assay_titles I think

shengchd (changes requested):
Looks good! have a minor request on changelog.

shengchd (dismissed):

PR #1296, opened 2025-01-30
IGVF-2260-filtered-file

meenakshikagda:


jenjou (commented):
Since the property already exists for Alignment File and Signal File, I wonder if it would be best to move it into mixins so the property can be shared by all 4 subclasses of File? 

I think this code works so it is not wrong, but wondering if we can make things more centralized.

jenjou (dismissed):

jenjou (approved):

PR #1297, opened 2025-01-30
IGVF-2304-spot-barcode

ian-whaling:


jenjou (commented):
Are we confident this can only be used as principal dimension, and never as secondary dim?

jenjou (dismissed):

PR #1301, opened 2025-02-03
IGVF-2308-add-readnames-audit

zhwshen:


src/igvfd/audit/file_set.py
```diff
@@ -536,3 +536,42 @@ def audit_MPRA_read_names(value, system):
...
+            "audit_category": "missing read names",
+            "audit_level": "ERROR"
+        }
+    ]
+    '''
+
+    object_type = space_in_words(value['@type'][0]).capitalize()
+    audit_message_missing_readnames = get_audit_message(audit_single_cell_read_names, index=0)
+    # Check read_names info on SeqFile objects from single cell MeaSet
+    if single_cell_check(system, value, 'Measurement set'):
```
ian-whaling, line 556: You should be able to pass in the object_type instead of 'Measurement set'.

src/igvfd/audit/file_set.py
```diff
@@ -536,3 +536,42 @@ def audit_MPRA_read_names(value, system):
...
+
+
+@audit_checker('MeasurementSet', frame='object')
+def audit_single_cell_read_names(value, system):
+    '''
+    [
+        {
+            "audit_description": "Sequence files are expected to have read names.",
+            "audit_category": "missing read names",
+            "audit_level": "ERROR"
```
ian-whaling, line 548: NOT_COMPLIANT, not ERROR

src/igvfd/audit/file_set.py
```diff
@@ -536,3 +536,42 @@ def audit_MPRA_read_names(value, system):
...
+    audit_message_missing_readnames = get_audit_message(audit_single_cell_read_names, index=0)
+    # Check read_names info on SeqFile objects from single cell MeaSet
+    if single_cell_check(system, value, 'Measurement set'):
+        if 'files' in value:
+            file_info = {}
+            for file in value['files']:
+                if file.startswith('/sequence-files/'):
+                    sequence_file_object = system.get('request').embed(file)
+                    read_names = str(sequence_file_object.get('read_names', ''))
+                    file_info[file] = read_names
```
ian-whaling, line 563: This seems not necessary. All you have to do is collate the list of sequence files in the measurement set with no `read_names`. See the audit_MPRA_read_names audit for an example.

ian-whaling (changes requested):
Request to simplify audit + change category.

src/igvfd/audit/file_set.py
ian-whaling: missing properties are not ERROR since it is not incorrect metadata just incomplete.

src/igvfd/audit/file_set.py
zhwshen: Removed the object_type line because this audit is for only MeasurementSet, no need to calc that. Also, the single cell check function takes `Measurement set` while the object_type will return `Measurement Set`.

src/igvfd/audit/file_set.py
zhwshen: Simplified. Also removed checking if read_names will return an empty list (which it shouldn't). It is either an empty string or a valid array.

ian-whaling (approved):

PR #1304, opened 2025-02-04
IGVF-2314-adjust-audit-levels

ian-whaling:


ian-whaling:
Since coveralls won't load I can only presume it is complaining about that I added an additional check for checking if an analysis set is single cell and I did not create a test for it. I don't think it is necessary to add one ATP since most of the single cell audits only pertain to raw data.

hitz (dismissed):

PR #1307, opened 2025-02-05
IGVF-2298-tissue-summary

forresttanaka:


src/igvfd/types/sample.py
```diff
@@ -676,6 +676,28 @@ class Tissue(Biosample):
...
+        'x': {
+            'group_by': 'donors.strain',
+            'label': 'Strain'
+        },
+        'y': {
+            'group_by': ['sample_terms.term_name', 'sex'],
+            'label': 'Tissue'
+        }
+    }
+
```
keenangraham, line 700: These should go in search configs. Probably https://github.com/IGVF-DACC/igvfd/blob/dev/src/igvfd/searches/configs/tissue.py

Here is example of matrix config: https://github.com/ENCODE-DCC/encoded/blob/cc9be06035b9905a5f24b21fb97a1becf741cd1d/src/encoded/searches/configs.py#L66-L84

src/igvfd/search_views.py
```diff
@@ -360,6 +362,68 @@ def missing_matrix(context, request):
...
+            ),
+            TypeResponseField(
+                at_type=['TissueSummaryHomoSapiens']
+            ),
+            IDResponseField(),
+            SearchBaseResponseField(),
+            ContextResponseField(),
+            MissingMatrixWithFacetsResponseField(
+                default_item_types=DEFAULT_ITEM_TYPES,
+                matrix_definition_name='tissue_homo_sapiens',
```
keenangraham, line 383: I think remove this and add `config=['whatever-you-name-your-search-config']` (I think string instead of list also works), so the endpoint uses your search config with the matrix definition.

src/igvfd/search_views.py
```diff
@@ -360,6 +362,68 @@ def missing_matrix(context, request):
...
+            ),
+            TypeResponseField(
+                at_type=['TissueSummaryMusMusculus']
+            ),
+            IDResponseField(),
+            SearchBaseResponseField(),
+            ContextResponseField(),
+            MissingMatrixWithFacetsResponseField(
+                default_item_types=DEFAULT_ITEM_TYPES,
+                matrix_definition_name='tissue_mus_musculus',
```
keenangraham, line 414: Swap this with `config=...`, where search config contains matrix definition.

keenangraham (approved):
Great!

PR #1308, opened 2025-02-06
IGVF-2199-audit-control

ian-whaling:


src/igvfd/tests/test_audit_file_set.py
```diff
@@ -151,3 +151,49 @@ def test_audit_single_cell_read_names(testapp, measurement_set_one_onlist, seque
...
+    testapp,
+    measurement_set_mpra,
+    construct_library_set_reporter,
+    analysis_set_base
+):
+    res = testapp.get(measurement_set_mpra['@id'] + '@@audit')
+    assert all(
+        error['category'] != 'missing control for'
+        for error in res.json['audit'].get('NOT_COMPLIANT', [])
+    )
```
jenjou, line 166: I think it would make sense to add another assertion here that there is no "missing control type" audit present either.

src/igvfd/tests/test_audit_file_set.py
```diff
@@ -151,3 +151,49 @@ def test_audit_single_cell_read_names(testapp, measurement_set_one_onlist, seque
...
+        measurement_set_mpra['@id'],
+        {
+            'control_type': 'low FACS signal'
+        }
+    )
+    res = testapp.get(measurement_set_mpra['@id'] + '@@audit')
+    assert all(
+        error['category'] != 'missing control type'
+        for error in res.json['audit'].get('NOT_COMPLIANT', [])
+    )
```
jenjou, line 188: Could also assert there's no missing control for.

jenjou (changes requested):
Requesting a couple more assertions in the test but looks ok otherwise.

src/igvfd/tests/test_audit_file_set.py
ian-whaling: done

src/igvfd/tests/test_audit_file_set.py
ian-whaling: done

jenjou (dismissed):

PR #1311, opened 2025-02-08
IGVF-2310-biosample-qualifiers

meenakshikagda:


src/igvfd/upgrade/sample.py
```diff
@@ -513,3 +513,13 @@ def in_vitro_system_23_24(value, system):
             f' Please make sure it is correct before removing the notes.'
         )
         value['notes'] = notes.strip()
+
+
+@upgrade_step('primary_cell', '21', '22')
+@upgrade_step('in_vitro_system', '24', '25')
+def sample_21_22(value, system):
+    # https://igvf.atlassian.net/browse/IGVF-2310
+    value['biosample_qualifiers'] = ['6 days calcified']
```
jenjou, line 522: I think there should be a check before this so only
```
if 'calcified' in value['biosample_qualifiers']:
value['biosample_qualifiers'].remove('calcified')
value['biosample_qualifiers'].append('6 days calcified')
```

Otherwise this will add 6 days calcified onto every sample

src/igvfd/upgrade/sample.py
```diff
@@ -513,3 +513,15 @@ def in_vitro_system_23_24(value, system):
             f' Please make sure it is correct before removing the notes.'
         )
         value['notes'] = notes.strip()
+
+
+@upgrade_step('primary_cell', '21', '22')
+@upgrade_step('in_vitro_system', '24', '25')
+def sample_21_22(value, system):
+    # https://igvf.atlassian.net/browse/IGVF-2310
+    if 'calcified' in value['biosample_qualifiers']:
```
jenjou, line 522: 'biosample_qualifiers' is not required so need to implement something in case it does not exist in the object.
Like: `if 'calcified' in value.get('biosample_qualifiers', [])`

jenjou (changes requested):
Just requesting one change to make sure upgrade doesn't fail on objects without biosample_qualifiers. The rest of the PR looks OK

jenjou (approved):

PR #1315, opened 2025-02-11
IGVF-2342-calc-mux-samp-class

zhwshen:


src/igvfd/types/sample.py
```diff
@@ -9,12 +9,18 @@
...
+)
 
 
-def collect_multiplexed_samples_prop(request, multiplexed_samples, property_name):
+def collect_multiplexed_samples_prop(request, multiplexed_samples, property_name, skip_calculated=True):
     property_set = set()
     for sample in multiplexed_samples:
-        sample_props = request.embed(sample, '@@object?skip_calculated=true')
+        if skip_calculated is False:
+            sample_props = request.embed(sample, '@@object')
```
ian-whaling, line 21: since you just want the classification from the underlying samples and not the other calc'd props you can do: @@object_with_select_calculated_properties?field=classifications

ian-whaling (dismissed):

PR #1332, opened 2025-02-21
IGVF-2374-gencode-v22

meenakshikagda:


ian-whaling (changes requested):
I think it would be better practice to add new inserts for any new enums you're adding as theoretically what you're adding could affect past inserts and by updating them you're losing that test for the past enum.

ian-whaling (approved):

PR #1333, opened 2025-02-21
IGVF-2365-definition-ontology

meenakshikagda:


mingjiecn (changes requested):
Can you also add some sample data that has comment?

mingjiecn (approved):

PR #1337, opened 2025-02-22
IGVF-2379-extend-gene-chrom

shengchd:


shengchd:
> Should the chromosome property under loci in mixins also be updated to support non-standard chromosomes?

yes I think it would make sense to update them , will do

shengchd:
Actually regex in mixins already allow non-standard chromosomes, I will just update the regex under genes and make it consistent with the one in mixins (^(chr[0-9A-Za-z**_**]+)$)

ian-whaling (approved):

PR #1345, opened 2025-02-26
IGVF-2357-ic-props-facets

forresttanaka:


src/igvfd/types/institutional_certificate.py
```diff
@@ -28,3 +28,18 @@ class InstitutionalCertificate(Item):
 
     set_status_up = []
     set_status_down = []
+
+    @calculated_property(schema={
+        'title': 'Data Use Limitation Summary',
+        'type': 'string',
+        'description': 'A unique identifier for the ontology term, reformatted from the original term ID.',
```
jenjou, line 35: Description sounds like another property, probably just a copy paste mistake?

forresttanaka: Yeah, I didn't even notice the description.

jenjou (changes requested):
Code looks fine except for the description of the new property, which needs to be updated.

src/igvfd/types/institutional_certificate.py
```diff
@@ -28,3 +28,18 @@ class InstitutionalCertificate(Item):
 
     set_status_up = []
     set_status_down = []
+
+    @calculated_property(schema={
+        'title': 'Data Use Limitation Summary',
+        'type': 'string',
+        'description': 'A combination of the data use limitation and its modifiers',
+        'notSubmittable': True,
+        'uniqueKey': True
```
jenjou, line 37: Sorry - missed this earlier. I think this "uniqueKey" can probably also be dropped.

forresttanaka: OK, that makes sense actually. I saw it in other calculated properties but probably misinterpreted.

PR #1351, opened 2025-02-28
IGVF-2418-fix-audit-transcript

ian-whaling:


jenjou (commented):
The code change here looks fine but since we are updating this audit - can we fix the typos on lines 351 and 357? In "skip calculated". It might not requesting the object as intended due to the typo.

jenjou (approved):

PR #1352, opened 2025-02-28
IGVF-2415-add-validate-onlist-on-seqspec

zhwshen:


src/igvfd/schemas/changelogs/configuration_file.md
```diff
@@ -2,6 +2,7 @@
 
 ### Minor changes since schema version 7
 
+* Add calculated property `validate_onlist_files`
```
jenjou, line 5: Minor request: can we add a period at the end of the entry?

src/igvfd/tests/test_types_file.py
```diff
@@ -460,3 +460,33 @@ def test_upload_credentials_allowed_when_status_is_preview(testapp, reference_fi
...
+
+
+def test_validate_onlist_files(testapp, configuration_file_json, configuration_file_seqspec, base_auxiliary_set, measurement_set_one_onlist):
+    # If a Config File is not a seqspec
+    res = testapp.get(configuration_file_json['@id'])
+    assert res.json.get('validate_onlist_files', '') is False
+
+    # IF a ConfigFile seqpec is not linked to a single cell measeurement set
+    res = testapp.get(configuration_file_seqspec['@id'])
+    assert res.json.get('validate_onlist_files', '') is False
```
jenjou, line 472: Since configuration_file_seqspec was not patched, this seems like the same assertion as line 468? Maybe there should be a patch before this or before line 468?

src/igvfd/types/file.py
```diff
@@ -755,6 +759,27 @@ def summary(self, content_type, seqspec_of=None):
             seqspec_of_formatted = f" of {', '.join(file_accessions)}"
         return f'{content_type}{seqspec_of_formatted}'
 
+    @calculated_property(
+        schema={
+            'title': 'Validate Onlist Files',
+            'type': 'boolean',
+            'description': 'Whether to checkfiles will validate the onlist files.',
```
jenjou, line 766: Minor edit: I think the word "to" is a typo here

jenjou (commented):
Mostly looks good but have one question on the test and a couple typos.

src/igvfd/tests/test_types_file.py
zhwshen: @jenjou 
The first config file is a json (not seqspec). To test that if a config file is not seqspec, it automatically gets a False for onlist validation. The second seqspec config file is one not linked to a single cell measurement set, that also gets an automatic False.

jenjou: Oops, sorry I missed that! I see now that they are different test fixtures.

jenjou (dismissed):

PR #1355, opened 2025-02-28
IGVF-2401-fix-audit-nucleic

meenakshikagda:


src/igvfd/types/sample.py
```diff
@@ -1038,7 +1038,10 @@ def sources(self, request, multiplexed_samples):
         }
     )
     def construct_library_sets(self, request, multiplexed_samples):
-        return collect_multiplexed_samples_prop(request, multiplexed_samples, 'construct_library_sets')
+        if collect_multiplexed_samples_prop(request, multiplexed_samples, 'construct_library_sets'):
```
ian-whaling, line 1041: This is calculating the property twice when you only need to do so once. Define it as a variable and if it has entries return it else return None

ian-whaling (dismissed):

ottojolanki (approved):

PR #1362, opened 2025-03-03
IGVF-2411-human-mouse-reference-schema

khzlin:


src/igvfd/schemas/changelogs/tabular_file.md
```diff
@@ -2,6 +2,15 @@
 
 ### Minor changes since schema version 13
 
+* Extend `transcriptome_annotation` enum list to include `GENCODE 28`.
+* Extend `transcriptome_annotation` enum list to include `GENCODE 25`.
+* Extend `transcriptome_annotation` enum list to include `GENCODE M17`.
+* Extend `transcriptome_annotation` enum list to include `GENCODE 28, GENCODE M17`.
+* Extend `assembly` enum list to include `hg19`.
+* Extend `assembly` enum list to include `Cast - GRCm39`.
```
jenjou, line 10: I don't think this was newly added, it just got moved lower in the list, so we don't need to include a new changelog entry for it

src/igvfd/schemas/reference_file.json
```diff
@@ -220,6 +220,9 @@
             "submissionExample": {
                 "appscript": "variants",
                 "igvf_utils": "variants"
+            },
+            "assembly": {
```
jenjou, line 224: I think this was accidentally added?

src/igvfd/schemas/reference_file.json
```diff
@@ -245,17 +248,6 @@
                 "igvf_utils": "vcf"
             }
         },
-        "assembly": {
```
jenjou, line 248: I think this part can be kept, and only remove the enum list in this.

src/igvfd/schemas/changelogs/reference_file.md
```diff
@@ -1,5 +1,13 @@
 ## Changelog for *`reference_file.json`*
 
+### Minor changes since schema version 16
```
jenjou, line 3: Include a blank line after a header

jenjou (changes requested):
See comments

src/igvfd/schemas/changelogs/tabular_file.md
```diff
@@ -2,6 +2,15 @@
 
 ### Minor changes since schema version 13
 
+* Extend `transcriptome_annotation` enum list to include `GENCODE 28`.
+* Extend `transcriptome_annotation` enum list to include `GENCODE 25`.
```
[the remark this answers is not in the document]
khzlin: done

src/igvfd/schemas/changelogs/reference_file.md
khzlin: done

src/igvfd/schemas/reference_file.json
khzlin: removed

khzlin: added back because your other comment said this part can be kept

src/igvfd/schemas/reference_file.json
khzlin: done

src/igvfd/schemas/changelogs/tabular_file.md
khzlin: done.

src/igvfd/schemas/reference_file.json
jenjou: Sorry for the confusion, but I think line 224 should not be here (it is within the `content_type` definition which does not make sense to me?). 

And line 248 could be restored depending on whether we want to have a special property description for Reference File. But if we just want the original description then we can keep that deleted.

jenjou (dismissed):

src/igvfd/schemas/reference_file.json
khzlin: done

jenjou (approved):

PR #1363, opened 2025-03-04
IGVF-2309-filtered-summary

meenakshikagda:


src/igvfd/types/file.py
```diff
@@ -629,13 +629,19 @@ def content_summary(self, principal_dimension, secondary_dimensions, content_typ
...
         file_set_object = request.embed(file_set, '@@object_with_select_calculated_properties?field=@type')
         predicted = None
         if 'PredictionSet' in file_set_object['@type']:
             predicted = 'predictive'
+        if filtered is True:
+            filtered_phrase = 'filtered'
+        elif filtered is False:
+            filtered_phrase = 'unfiltered'
+        else:
+            filtered_phrase = 'unfiltered'
```
jenjou, line 642: Since there are only two cases, I think this section can be simplified somewhat. Something like: 
```
filtered_phrase = 'unfiltered'
if filtered:
   filtered_phrase = 'filtered'
````
 This can also apply to the summary property for Tab File

jenjou (changes requested):
Small code suggestion but otherwise this should work.

src/igvfd/types/file.py
meenakshikagda: Updated

jenjou (dismissed):

jenjou (approved):
Requires a rebase

ottojolanki:
I'm rebasing. Is this ready for merge then?

PR #1371, opened 2025-03-06
IGVF-2426-remove-crispr-growth

meenakshikagda:


jenjou (changes requested):
The enum is also used by Assay Term in `preferred_assay_titles`, so there will need to be a similar upgrade, test, and changelog update.

src/igvfd/tests/test_upgrade_assay_term.py
```diff
@@ -60,3 +60,9 @@ def test_assay_term_upgrade_9_10(upgrader, assay_term_v9):
     value = upgrader.upgrade('assay_term', assay_term_v9, current_version='9', target_version='10')
     assert value['schema_version'] == '10'
     assert set(value.get('preferred_assay_titles')) == {'10x multiome with scMito-seq'}
+
+
+def test_assay_term_upgrade_9_10(upgrader, assay_term_v10):
```
jenjou, line 65: The name of this test should be updated and I think the fixture "assay_term_v10" is missing from the changes (it seems to be causing the test failure)

PR #1384, opened 2025-03-13
IGVF-2417-ic-matching-audit

jenjou:


src/igvfd/audit/measurement_set.py
```diff
@@ -266,6 +266,77 @@ def audit_missing_institutional_certification(value, system):
...
+        yield AuditFailure(
+            audit_message_mixed_files.get('audit_category', ''),
+            f'{detail} {audit_message_mixed_files.get("audit_description", "")}',
+            level=audit_message_mixed_files.get('audit_level', '')
+        )
+    elif True in files_by_access and False not in files_by_access:
+        controlled_access_of_files_phrase = 'controlled access'
+    elif True not in files_by_access and False in files_by_access:
+        controlled_access_of_files_phrase = 'uncontrolled access'
+
```
shengchd, line 315: I was looking at if controlled_access is a required boolean field for all files. In index file and model files they are optional, config/matrix/signal files they don't have that property. So theoretically could be a case where all files don't have `controlled_access` and the key in `files_by_access` is only an empty string -> `controlled_access_of_files_phrase` will not be assigned before it might be called below. This seems to be an invalid case for meaSet, but I think you could skip files without `controlled_access` in files_by_access to be safe.

shengchd: Also I feel maybe we should make controlled_access as a required field for index file, or be calculated from its derived_from file, which could be alignment file.

src/igvfd/tests/test_audit_measurement_set.py
```diff
@@ -421,6 +421,48 @@ def test_audit_missing_institutional_certification(
...
+            'file_set': measurement_set['@id']
+        }
+    )
+    res = testapp.get(measurement_set['@id'] + '@@audit')
+    assert any(
+        error['category'] == 'inconsistent controlled access'
+        for error in res.json['audit'].get('ERROR', [])
+    )
+
+
```
shengchd, line 465: can we also add a case where files have the same access status but it's different from IC?

shengchd (commented):
The code looks good, have a suggestion for completeness and a minor request on adding another test case.

src/igvfd/audit/measurement_set.py
jenjou: I wonder if that is possible to change since I am not sure we could upgrade Index File. I guess this could be a future ticket.

src/igvfd/audit/measurement_set.py
```diff
@@ -266,6 +266,79 @@ def audit_missing_institutional_certification(value, system):
...
+            f'submitted as uncontrolled access.'
+        )
+        yield AuditFailure(
+            audit_message_mixed_files.get('audit_category', ''),
+            f'{detail} {audit_message_mixed_files.get("audit_description", "")}',
+            level=audit_message_mixed_files.get('audit_level', '')
+        )
+    elif True in files_by_access and False not in files_by_access:
+        controlled_access_of_files_phrase = 'controlled access'
+    elif True not in files_by_access and False in files_by_access:
```
shengchd, line 314: I would change the line here just to be 'else:', so it will cover the edge case if there's no key in files_by_access after skipping the none cases (sorry my last comment didn't mention that)

shengchd (dismissed):

src/igvfd/audit/file_set.py
```diff
@@ -660,3 +660,81 @@ def audit_control_for_control_type(value, system):
...
+            sample, '@@object_with_select_calculated_properties?field=institutional_certificates&field=@id')
+        for ic in sample_object.get('institutional_certificates'):
+            ic_object = system.get('request').embed(ic, '@@object_with_select_calculated_properties?field=@id')
+            if any(ic_object.get('controlled_access') != access for access in files_by_access):
+                if ic_object.get('controlled_access') is True:
+                    controlled_access_of_ic_phrase = 'controlled access'
+                else:
+                    controlled_access_of_ic_phrase = 'uncontrolled access'
+                detail = (
+                    f'Measurement set {audit_link(path_to_text(value["@id"]), value["@id"])} '
```
shengchd, line 730: would need adjust 'Measurement set' from object type to include analysis set

shengchd (changes requested):
a minor bug in error detail phrase

shengchd (approved):

PR #1391, opened 2025-03-18
IGVF-2428-strand-specificity

meenakshikagda:


src/igvfd/audit/measurement_set.py
```diff
@@ -492,6 +492,35 @@ def audit_missing_auxiliary_set(value, system):
...
+            "audit_category": "missing strand specificity",
+            "audit_level": "WARNING"
+        }
+    ]
+    '''
+    audit_message_strand_specificity = get_audit_message(audit_missing_strand_specificity)
+
+    expected_assays = {'Perturb-seq', 'scCRISPR screen', 'TAP-seq', 'CERES-seq'}
+    preferred_assay_title = value.get('preferred_assay_title')
+    strand_specificity = value.get('strand_specificity')
```
jenjou, line 510: Although this is okay I would recommend to define the default value like `value.get('strand_specificity', None)` so that it is explicit it will return "none".

src/igvfd/audit/measurement_set.py
```diff
@@ -492,6 +492,35 @@ def audit_missing_auxiliary_set(value, system):
                 yield AuditFailure(audit_message.get('audit_category', ''), f'{detail} {audit_message.get("audit_description", "")}', level=audit_message.get('audit_level', ''))
 
 
+@audit_checker('MeasurementSet', frame='embedded')
```
jenjou, line 495: It doesn't seem like this needs frame=embedded - was the test failing if it was not embedded?

src/igvfd/schemas/measurement_set.json
```diff
@@ -302,6 +302,20 @@
                 "igvf_utils": "direct RNA, exome capture"
             }
         },
+        "strand_specificity": {
+            "title": "Strand specificity",
```
jenjou, line 306: Specificity should be capitalized.

src/igvfd/schemas/measurement_set.json
```diff
@@ -302,6 +302,20 @@
...
+            "description": "The strand-specificty of the sequencing results within Perturb-seq, scCRISPR screen, TAP-seq, and CERES-seq assays.",
```
jenjou, line 307: Small typo, "specificity"

src/igvfd/audit/measurement_set.py
meenakshikagda: I should try that.

src/igvfd/schemas/measurement_set.json
meenakshikagda: Only in the title you mean right? Or should I update it in the description as well?

jenjou (approved):

PR #1393, opened 2025-03-18
Update MPRA quality metrics

visze:
Changes to clarify values, use fractions instead of percentage and fix typos.

(see  last FG meeting)


keenangraham:
Thanks, done in https://github.com/IGVF-DACC/igvfd/pull/1406

PR #1431, opened 2025-03-31
IGVF-2099-audit-float-sample

ian-whaling:


src/igvfd/tests/test_audit_sample.py
```diff
@@ -163,3 +163,29 @@ def test_audit_missing_nucleic_acid_delivery(
         error['category'] != 'missing nucleic acid delivery'
         for error in res.json['audit'].get('WARNING', [])
     )
+
+
+def test_audit_missing_nucleic_acid_delivery(
```
jenjou, line 168: Test name should be updated

jenjou (changes requested):
Code looks okay, just requesting change to the test name.

jenjou (dismissed):
Will need rebase again, probably after the last few tickets from the current release get in.

PR #1447, opened 2025-04-07
IGVF-2029-rm-dup-props-in-images

zhwshen:


jenjou (changes requested):
It looks like `thumb_nail` is a calculated property, therefore I think we don't actually need to upgrade the objects. 

It should still have a changelog entry but it can be placed in the existing "minor changes since..." section.

src/igvfd/schemas/changelogs/image.md
```diff
@@ -2,6 +2,7 @@
 
 ### Minor changes since schema version 4
 
+* Remove `thumb_nail`.
```
jenjou, line 5: Can we say "Remove calculated property thumb_nail"? I think this will be more consistent with how we have recorded the addition of new calculated properties e.g. "Add calculated property controlled_access."

jenjou (approved):

PR #1453, opened 2025-04-09
IGVF-2548-add-assessed-genes-to-models

zhwshen:


src/igvfd/types/file_set.py
```diff
@@ -1278,11 +1310,19 @@ class PredictionSet(FileSet):
...
         scope_phrase = ''
         if scope:
             scope_phrase = f' on scope of {scope}'
-        return f'{file_set_type} prediction{scope_phrase}'
+        # Get assessed genes info
+        assessed_genes_phrase = get_assessed_gene_phrase(request, assessed_genes)
+        # Final summary
+        return ' '.join(filter(None, [
+            file_set_type,
+            f'prediction{scope_phrase}',
```
ian-whaling, line 1323: I think this does not filter out scope_phrase properly if it is None

src/igvfd/tests/test_types_prediction_set.py
```diff
@@ -0,0 +1,44 @@
...
+                    gene_CRLF2_par_y['@id'],
+                    gene_CD1E['@id'],
+                    gene_TAB3_AS1['@id'],
+                    gene_MAGOH2P['@id']
+                ]
+        }
+    )
+    res = testapp.get(base_prediction_set['@id'])
+    print('CHECK CHECK')
+    print(res.json.get('summary'))
```
ian-whaling, line 25: remove print statements

src/igvfd/types/file_set.py
ian-whaling: Oh I see that scope_phrase is actually defined as an empty string above

ian-whaling (changes requested):
remove print statements

ian-whaling (dismissed):

PR #1466, opened 2025-04-15
IGVF-2588-audit-anvil

ian-whaling:


src/igvfd/audit/file.py
```diff
@@ -112,3 +112,34 @@ def audit_file_no_file_format_specifications(value, system):
...
+@audit_checker('TabularFile', frame='object')
+@audit_checker('ReferenceFile', frame='object')
+@audit_checker('ModelFile', frame='object')
+@audit_checker('SequenceFile', frame='object')
+def audit_file_missing_anvil_url(value, system):
+    '''
+    [
+        {
+            "audit_description": "Released and archived files that are restricted access are expected to be loaded in anvil.",
+            "audit_category": "missing anvil url",
```
keenangraham, line 128: Wondering if there will be cases in future where we move files that aren't controlled access to Anvil.

src/igvfd/audit/file.py
```diff
@@ -112,3 +112,34 @@ def audit_file_no_file_format_specifications(value, system):
...
+            "audit_level": "INTERNAL_ACTION"
+        }
+    ]
+    '''
+    object_type = space_in_words(value['@type'][0]).capitalize()
+    audit_message = get_audit_message(audit_file_missing_anvil_url, index=0)
+    if value.get('status') in ['released', 'archived'] and value.get('controlled_access', False) and not (value.get('anvil_url')):
```
keenangraham, line 135: `(value.get('anvil_url'))` Don't think parens necessary.

src/igvfd/audit/file.py
```diff
@@ -112,3 +112,34 @@ def audit_file_no_file_format_specifications(value, system):
...
+        file_set_object = system.get('request').embed(
+            value.get('file_set'), '@@object_with_select_calculated_properties?field=data_use_limitation_summaries')
+        if file_set_object.get('status') in ['released', 'archived']:
+            data_use_limitation_summaries = file_set_object.get('data_use_limitation_summaries', '')
+            if data_use_limitation_summaries and data_use_limitation_summaries != ['no certificate']:
```
keenangraham, line 140: Controlled-access file but no certificates? Seems a little shaky to rely on string value from summary.

src/igvfd/audit/file.py
```diff
@@ -112,3 +112,34 @@ def audit_file_no_file_format_specifications(value, system):
...
+                detail = (
+                    f'{object_type} {audit_link(path_to_text(value["@id"]), value["@id"])} '
```
keenangraham, line 142: Think would be nice to have clear indication of which Anvil project a file needs to be moved to based on its DUL code or data use summary etc. But maybe that's still being worked out.

keenangraham (approved):

src/igvfd/audit/file.py
ian-whaling: There are a series of audits that ensure a certificate is submitted for file sets that require one prior to the release of either the file or file set so this will likely never be encountered.

src/igvfd/audit/file.py
ian-whaling: I can add it to the detail but for your usage I think the facet will be more useful. The detail is just displayed in the audit

src/igvfd/audit/file.py
ian-whaling: fixed

src/igvfd/audit/file.py
ian-whaling: I guess if we introduce another workspace for hosting restricted data I can update this to be more general. The audit would have to be updated anyways coz the property anvil_url is specific to that workspace atm.

cricketsloan: Do you mean that the property has aregex that specifies the url prefix?

ian-whaling: There is no regex, but to submit anything other than anvil to anvil_url would be weird and if there's another workspace I think that would necessitate either another property or to rename anvil_url to something more general.

src/igvfd/audit/file.py
```diff
@@ -112,3 +112,35 @@ def audit_file_no_file_format_specifications(value, system):
...
+        {
+            "audit_description": "Released and archived files that are restricted access are expected to be loaded in anvil.",
+            "audit_category": "missing anvil url",
+            "audit_level": "INTERNAL_ACTION"
+        }
+    ]
+    '''
+    object_type = space_in_words(value['@type'][0]).capitalize()
+    audit_message = get_audit_message(audit_file_missing_anvil_url, index=0)
+    if value.get('status') in ['released', 'archived'] and value.get('controlled_access', False) and not value.get('anvil_url'):
```
cricketsloan, line 135: Do we want to put archived data into Anvil?  I had not thought about it, i was focused on released state.

src/igvfd/audit/file.py
cricketsloan: @keenangraham I think the way to think about the "no certificate" state is that if we have said that the file is controlled, but a certificate does not exist, we don't want to put this into anvil yet as something is wrong

src/igvfd/audit/file.py
ian-whaling: archived is functionally the same as released so I see no reason why it should be deposited to anvil (and therefore made accessible).

ian-whaling: should not be*

src/igvfd/audit/file.py
cricketsloan: There will be several end points at anvil.  Examples
https://anvil.terra.bio/#workspaces/anvil-datastorage/AnVIL_IGVF_GRU
https://anvil.terra.bio/#workspaces/anvil-datastorage/AnVIL_IGVF_GRU-PUB

src/igvfd/audit/file.py
cricketsloan: I agree that the detail will be nice but the facet will be more useful

src/igvfd/audit/file.py
cricketsloan: My thought had been that archived means that it is no longer interesting so I would wonder why we would want to go to the trouble of propogating it to another site.    I had even wondered if we should remove archived data from the Anvil site as we have no way to indicate that it is different from released data

keenangraham (approved):

PR #1468, opened 2025-04-15
IGVF-2581-adjust-seqspec-audits

zhwshen:


src/igvfd/audit/file_set.py
```diff
@@ -138,11 +140,12 @@ def audit_files_associated_with_incorrect_fileset(value, system):
     '''
     object_type = space_in_words(value['@type'][0]).capitalize()
     audit_message = get_audit_message(audit_files_associated_with_incorrect_fileset)
+    # Non-sc seqspec submitted through measurement sets should still be audited like single cells
```
ian-whaling, line 143: file sets not measurement sets?

src/igvfd/audit/file_set.py
```diff
@@ -138,11 +140,12 @@ def audit_files_associated_with_incorrect_fileset(value, system):
...
     if 'files' in value:
         for file in value['files']:
+            # Audit 1: Seqspec have seqspec_of files taht are not in the same file set as the seqspec
```
ian-whaling, line 146: that*

src/igvfd/audit/file_set.py
```diff
@@ -217,7 +224,12 @@ def audit_inconsistent_seqspec(value, system):
                 )
                 yield AuditFailure(audit_message.get('audit_category', ''), f'{detail} {audit_message.get("audit_description", "")}', level=audit_message.get('audit_level', ''))
 
-        seqspec_to_sequence = {}
+        # Audit 2: If the same seqspec is linked to different sequencing sets, flag it
+        # now limit it to single cell only
+        if not single_cell_check(system, value, object_type):
```
ian-whaling, line 229: If we allow someone to submit their seqspec to a curated set for non-sc data then won't the missing related files audit above still trigger for the measurement sets which have sequence files which link to this seqspec located in the curated set but not in the same measurement set?

zhwshen: @ian-whaling The current consensus is that non-single cell seqspec can be submitted through either 1) Document, completely different audit, 2) MeasurementSet/AuxiliarySet/CLS, then the file set matching audit remains. They shouldn't submit through Curated Set.

ian-whaling (approved):

PR #1471, opened 2025-04-16
IGVF-2632-validation-exempt

meenakshikagda:


src/igvfd/types/file.py
```diff
@@ -1147,7 +1147,7 @@ def validate_bucket_location(request, properties, bucket):
 )
 def post_upload(context, request):
     properties = context.upgrade_properties()
-    if properties['upload_status'] == 'validated':
+    if properties['upload_status'] in ['validated', 'validation exempted']:
```
ottojolanki, line 1150: I cannot imagine any other statuses, but wondering if explicitly spelling out the allowed ones (by saying `not in ['pending', 'invalidated', 'file not found']` which right now is of course equivalent but would not change in case of added statuses) would be more future proof?

ottojolanki (changes requested):
I think being explicit is more future proof?

src/igvfd/types/file.py
```diff
@@ -1147,9 +1147,9 @@ def validate_bucket_location(request, properties, bucket):
 )
 def post_upload(context, request):
     properties = context.upgrade_properties()
-    if properties['upload_status'] == 'validated':
+    if properties['upload_status'] not in ['pending', 'file not found', 'invalidated']:
         raise HTTPForbidden(
-            'Unable to issue new credentials when uploading_status is validated'
+            'Unable to issue new credentials when uploading_status is validated or validation exempted'
```
ottojolanki, line 1152: Now that we are touching this, could you please fix the old typo in the error message and change `uploading_status` to the correct `upload_status`, please?

ottojolanki (dismissed):

PR #1486, opened 2025-04-21
IGVF-2625-sc-multiome-nt

jenjou:


src/igvfd/schemas/measurement_set.json
```diff
@@ -107,6 +107,7 @@
                 "snRNA-seq",
                 "scNT-seq",
                 "scNT-seq2",
+                "scMultiome-NT-seq",
```
shengchd, line 110: can we also add an enum description below?

shengchd (commented):
looks good, just a small request on adding enum description

PR #1491, opened 2025-04-22
IGVF-2535-relax-indexfile-audit

khzlin:


src/igvfd/tests/test_audit_index_file.py
```diff
@@ -18,3 +18,14 @@ def test_audit_index_files(testapp, reference_file, index_file_tbi):
...
         for audit in res.json['audit'].get('ERROR', {})
     )
+    testapp.patch_json(
+        index_file_tbi['@id'],
+        {
+            'derived_from': [tabular_file_bed['@id']]
+        }
+    )
+    res = testapp.get(index_file_tbi['@id'] + '@@audit')
+    assert not any(
```
jenjou, line 28: Usually we write this kind of assertion like the case in line 6 above. Is there a reason to use not any() vs all()?

khzlin: fixing this to be
```assert all(
        audit['category'] != 'unexpected indexed file'
        for audit in res.json['audit'].get('ERROR', {})
    )```

jenjou (approved):

PR #1496, opened 2025-04-23
IGVF-2672-add-barcode-replace-file-audit

zhwshen:


src/igvfd/audit/analysis_set.py
```diff
@@ -340,3 +340,36 @@ def audit_missing_transcriptome(value, system):
...
         )
         yield AuditFailure(audit_message.get('audit_category', ''), f'{detail} {audit_message.get("audit_description", "")}', level=audit_message.get('audit_level', ''))
+
+
+@audit_checker('AnalysisSet', frame='object')
+def audit_multiple_barcode_replacement_files_in_input(value, system):
+    '''
+    [
+        {
+            "audit_description": "All input Parse SPLiTseq measurement sets linked to the same analysis set are expected to have the same barcode replacement file.",
```
jenjou, line 350: The assay title includes the hyphen in SPLiT-seq so it would make sense to use that anywhere it appears in the audit description or audit detail.

jenjou (dismissed):

PR #1498, opened 2025-04-23
IGVF-2634-superstarr-summary

ian-whaling:


src/igvfd/audit/construct_library_set.py
```diff
@@ -11,14 +11,11 @@
...
     assay_terms = set()
-    for sample in value.get('applied_to_samples', []):
-        sample_object = system.get('request').embed(
-            sample + '@@object_with_select_calculated_properties?field=file_sets')
-        file_sets = sample_object.get('file_sets', [])
-        for file_set in file_sets:
-            if file_set.startswith('/measurement-sets/'):
-                input_file_set_object = system.get('request').embed(file_set + '@@object?skip_calculated=true')
-                assay_terms.add(input_file_set_object.get('assay_term'))
+    file_sets = value.get('file_sets', [])
```
khzlin, line 14: value here is CLS, not sure why you file_sets = value.get('file_sets', []) . CLS doesn't have file_sets

khzlin (dismissed):

khzlin:
the code looks good, but i have a thought:

for `file_sets` on CLS, why not name it `measurement_sets` since MeasurementSets have `construct_library_sets` and `auxiliary_sets` and would be good to keep consistent.

ian-whaling:
> the code looks good, but i have a thought:
> 
> for `file_sets` on CLS, why not name it `measurement_sets` since MeasurementSets have `construct_library_sets` and `auxiliary_sets` and would be good to keep consistent. Same for AuxiliarySets.. it has `construct_library_sets` and `measurement_sets`.

All file_sets have construct_library_sets coz all file_sets can have samples and therefore construct library sets.

khzlin (approved):

PR #1499, opened 2025-04-23
IGVF-2644-aset-summary-updates

jenjou:


src/igvfd/searches/configs/analysis_set.py
```diff
@@ -137,6 +137,9 @@ def analysis_set():
             },
             'functional_assay_mechanisms.term_name': {
                 'title': 'Functional Assay Mechanisms'
+            },
+            'files.content_type': {
+                'title': 'File Content Type'
```
ian-whaling, line 142: title should match the facet above: 'File Type'

ian-whaling (changes requested):
minor comment on column name

ian-whaling (dismissed):

PR #1500, opened 2025-04-23
IGVF-2658-desired-count

ottojolanki:


cdk/infrastructure/constructs/backend.py
```diff
@@ -218,6 +216,15 @@ def _define_fargate_service(self) -> None:
             redirect_http=True,
         )
 
+        # For existing services, this prevents the desiredCount from being overwritten
```
keenangraham, line 219: Are you sure this is needed? Should be in it's own method if so.

keenangraham (approved):

PR #1505, opened 2025-04-28
IGVF-2362-remove-facet-groups

ian-whaling:


shengchd (commented):
looks good, we have facet_groups in another two places: search_veiws.py and test_search_views.py, might want to get rid of them as well?

shengchd (approved):

PR #1533, opened 2025-05-10
IGVF-2683-analysis-summary-SGE

shengchd:


src/igvfd/types/file_set.py
```diff
@@ -74,8 +74,30 @@ def get_cls_phrase(cls_set):
...
     elif len(cls_phrases) > 2:
-        cls_phrase = ', '.join(cls_phrases[:-1]) + ', and ' + cls_phrases[-1]
+        if cls_phrases[0].startswith('an editing template library'):
+            common_phrase = cls_phrases[0].split(' in ')[0]
+            # only condense sentence if all CLS have the same selection criteria (i.e. sequence variants)
+            if all(cls_phrase.startswith(common_phrase) for cls_phrase in cls_phrases):
+                targeton_dict = {}
+                for cls_phrase in cls_phrases:
+                    # e.g. editing template library targeting sequence variants in exon6B of BARD1
+                    gene = cls_phrase.split(' ')[-1]
```
shengchd, line 86: I'm not sure if it's a good idea to get gene/targeton from CLS summary instead of real properties, but feel it's fine here if only for SGE case (assuming the CLS summary code is stable).

ian-whaling: It is stable for now, your test would be enough to catch if it is changed.

PR #1535, opened 2025-05-12
IGVF-2085-boolean-property-defaults

khzlin:


src/igvfd/schemas/file.json
```diff
@@ -31,20 +31,28 @@
...
+            "comment": "If derived_manually is true, analysis_step_version is not required.",
+            "oneOf": [
+                {
+                    "properties": {
+                        "derived_manually": {
+                            "const": true
+                        }
+                    },
+                    "not": {
+                        "required": [
```
shengchd, line 43: is it intended here if dervied_manually = true, then you can't link analysis_step_version to the file?

shengchd (changes requested):
Overall look good! Had a question, also need to update the schema version in schemas/tissue.json.

src/igvfd/schemas/file.json
khzlin: yes, since it is derived_manually, no softwares/softare version/analysis step are used so no analysis step version is needed.

shengchd (approved):

shengchd (dismissed):

shengchd (dismissed):

PR #1543, opened 2025-05-19
IGVF-2319-aset-summary-cls-only

jenjou:


src/igvfd/types/file_set.py
```diff
@@ -274,15 +273,28 @@ def input_for(self, request, input_for):
...
+            all_inputs_cls = True
+            cls_in_input_file_sets = set()
+            if input_file_sets:
+                for input_set in input_file_sets:
+                    if input_set.startswith('/construct-library-sets/'):
+                        cls_in_input_file_sets.add(input_set)
+                    else:
+                        all_inputs_cls = False
+            if all_inputs_cls:
+                construct_library_sets = construct_library_sets | cls_in_input_file_sets
```
shengchd, line 297: this line can be simplified: construct_library_sets = cls_in_input_file_sets, but it works fine.

shengchd (commented):
The code looks good, have a minor suggestion. 
Should we also revise the definition of that calculated property since now it can be assigned under analysis set without samples?

> The construct library sets associated with the samples of this file set. (line 264)

shengchd (approved):

shengchd (approved):

PR #1571, opened 2025-06-05
IGVF-2751-cell-painting-audit

khzlin:


src/igvfd/audit/file_set.py
```diff
@@ -91,13 +91,28 @@ def audit_no_files(value, system):
     ]
     '''
     object_type = space_in_words(value['@type'][0]).capitalize()
+    preferred_assay_title = value.get('preferred_assay_title', '')
+
+    # Skip audit for specific Measurement set + assay combinations
+    skip_assays = [
+        'Variant painting via immunostaining',
+        'Variant painting via fluorescence',
+        'Cell Painting'
```
shengchd, line 100: 'P'ainting needs to be in lower case

src/igvfd/tests/test_audit_file_set.py
```diff
@@ -15,8 +15,17 @@ def test_audit_missing_files(
...
+    res = testapp.get(measurement_set_no_files['@id'] + '@@audit')
+    assert all(
+        error['category'] != 'missing files'
+        for error in res.json['audit'].get('NOT_COMPLIANT', [])
+    )
+    testapp.patch_json(
+        measurement_set_no_files['@id'],
         {
-            'file_set': measurement_set_no_files['@id']
+            'preferred_assay_title': 'Cell painting'
```
shengchd, line 28: surprised the test passed here - think `measurement_set_no_files` was patched with reference_file above, so it didn't really test what you intended to do here.

shengchd (changes requested):
Have a typo and test to fix, otherwise it's good

src/igvfd/audit/file_set.py
khzlin: done

PR #1580, opened 2025-06-11
IGVF-2735-rm-workflow

zhwshen:


src/igvfd/types/analysis_step.py
```diff
@@ -47,6 +46,49 @@ def unique_keys(self, properties):
         keys.setdefault('analysis_step:name', []).append(self._name(properties))
         return keys
 
+    @calculated_property(schema={
```
jenjou, line 49: I think in this section, you will require `define=True`, so that a calculated property can be used by another calculated property (like how "workflows" below is relying on this calculated "analysis_step_versions")

src/igvfd/tests/test_types_analysis_set.py
```diff
@@ -458,14 +458,13 @@ def test_functional_assay_mechanisms(testapp, analysis_set_base, measurement_set
...
-            'file_set': analysis_set_with_workflow['@id']
+            'file_set': analysis_set_base['@id'],
+            'analysis_step_version': analysis_step_version_3['@id']
         }
     )
-    res = testapp.get(analysis_set_with_workflow['@id'])
-    assert set([workflow['@id'] for workflow in res.json.get('workflows')]
-               ) == {'/workflows/IGVFWF0000WRKF/'}
+    res = testapp.get(analysis_set_base['@id'])
+    assert sorted(set([item['@id'] for item in res.json.get('workflows')])) == sorted(set([base_workflow_3['@id']]))
```
jenjou, line 470: We can keep this as is, but I just note that the `sorted(set(...))` seems a bit over-robust here for a list of one item. But it is not technically wrong so it's okay to leave it like this.

src/igvfd/types/file.py
```diff
@@ -290,24 +290,32 @@ def assay_titles(self, request, file_set):
...
+            'title': 'Workflows',
             'description': 'The workflow used to produce this file.',
-            'type': 'string',
-            'linkTo': 'Workflow',
+            'type': 'array',
+            'minItems': 1,
+            'uniqueItems': True,
+            'items': {
+                'title': 'Workflow',
+                'type': 'string',
```
jenjou, line 303: This will also cause the bug which will be resolved by IGVF-2831, in the mean time, we should keep this matching other arrays, like `['string', 'object']`

jenjou (approved):

src/igvfd/types/analysis_step_version.py
```diff
@@ -31,28 +31,28 @@ class AnalysisStepVersion(Item):
     embedded_with_frame = [
         Path('award', include=['@id', 'component']),
         Path('lab', include=['@id', 'title']),
-        Path('analysis_step', include=['@id', 'name', 'status', 'title', 'workflow']),
-        Path('analysis_step.workflow', include=['@id', 'accession', 'name', 'status']),
+        Path('analysis_step', include=['@id', 'name', 'status', 'title']),
```
ian-whaling, line 34: name is no longer a property of analysis step

src/igvfd/types/workflow.py
```diff
@@ -22,37 +22,20 @@ class Workflow(Item):
...
         Path('lab', include=['@id', 'title']),
         Path('submitted_by', include=['@id', 'title']),
         Path('standards_page', include=['@id', 'title', 'status']),
         Path('publications', include=['@id', 'publication_identifiers', 'status']),
-        Path('analysis_steps.analysis_step_versions.software_versions.software',
-             include=['@id', 'title', 'name', 'analysis_step_types', 'output_content_types', 'input_content_types', 'analysis_step_versions', 'software_versions', 'software', 'status'])
+        Path('analysis_step_versions.software_versions.software',
+             include=['@id', 'title', 'name', 'software_versions', 'software', 'status']),
+        Path('analysis_step_versions.analysis_step',
+             include=['@id', 'title', 'analysis_step_types', 'output_content_types', 'input_content_types', 'status'])
```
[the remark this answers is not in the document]
zhwshen: @keenangraham 
1. `anaysis_step_versions` is now a submittable property on Workflow object. Does that need to be embedded as a separate `Path('analysis_step_versions`, include=['software_versions', 'analysis_step'])`?
2. Is `Path(analysis_step_versions.analysis_step)` not sufficient?

src/igvfd/types/analysis_step.py
```diff
@@ -42,45 +41,19 @@ class AnalysisStep(Item):
...
-    })
-    def name(self):
-        return self.__name__
-
-    @property
-    def __name__(self):
-        properties = self.upgrade_properties()
-        return self._name(properties)
-
-    def _name(self, properties):
```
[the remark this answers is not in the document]
zhwshen: How do I do this?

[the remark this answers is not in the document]
keenangraham: Make new branch based off of `dev`. Spin up demo from that branch. Wait for it to deploy and index. Make sure there are some AnalysisStep objects on demo to tests. Then pull this code over to the branch and push. This should  allow you to see how deploying this code to running application behaves.

keenangraham:
Replaced with https://github.com/IGVF-DACC/igvfd/pull/1662

PR #1589, opened 2025-06-16
IGVF-2811-ethnicity-not-hispanic

khzlin:


jenjou (dismissed):
There was another ticket adding ethnicities, so a heads up this will have to be rebased soon.

PR #1594, opened 2025-06-18
IGVF-2833-control-types

ian-whaling:


hitz (approved):
this looks OK.  I have some concern about lower test coverage (although I think it's just because you added more lines).  You could maybe add back one of the tests with a single control type?

hitz (dismissed):

PR #1598, opened 2025-06-20
IGVF-2805-tissue-organ-part-2

jenjou:


src/igvfd/types/sample.py
```diff
@@ -373,6 +373,7 @@ def summary(self, request, sample_terms, donors, sex, age, age_units=None, modif
         term_object = request.embed(sample_terms[0], '@@object?skip_calculated=true')
         term_name = term_object.get('term_name')
         biosample_type = self.item_type
+        biosample_subschemas = ['primary_cell', 'in_vitro_system', 'tissue', 'whole_organism']
```
jenjou, line 376: Just to clarify: the reason "tissue" is not changed here is because this is a list of the schema names, and the schema name is going to remain as "tissue" (only the classifications changed or various titles).

zhwshen (approved):

PR #1599, opened 2025-06-23
IGVF-2822-matrix-transcriptome

shengchd:


src/igvfd/schemas/file.json
```diff
@@ -154,7 +154,7 @@
             }
         },
         "transcriptome_annotation": {
-            "comment": "The assembly shoud be consistent with annotation versions of respective species. Annotation version should not be submitted for custom assemblies.",
+            "comment": "The assembly shoud be consistent with annotation versions of respective species. Annotation version should not be submitted for custom assemblies. This dependency is not relevant for matrix files.",
```
shengchd, line 157: I was debating if this part should be moved to shared_dependencies.json instead, ended up with leaving a comment here for now since we have another ticket investigating all files on these two properties.

src/igvfd/types/file.py
```diff
@@ -680,6 +680,48 @@ def summary(self, request, content_summary, file_set, filtered=None, analysis_st
...
+            'description': 'The annotation and version of the reference resource.',
+            'notSubmittable': True
+        }
+    )
+    def transcriptome_annotation(self, request, reference_files):
+        transcriptome_annotation_set = set()
+        transcriptome_annotation = ''
+        for ref_file in reference_files:
+            ref_file_object = request.embed(ref_file, '@@object?skip_calculated=true')
+            if ref_file_object['content_type'] == 'transcriptome reference':
```
zhwshen, line 696: The content type should be transcriptome reference OR transcriptome index. Single cell uniform pipeline outputs RNA matrix files with only 1 reference that is a transcriptome index

src/igvfd/types/file.py
```diff
@@ -680,6 +680,48 @@ def summary(self, request, content_summary, file_set, filtered=None, analysis_st
...
+            'description': 'The assembly associated with the matrix file.',
+            'notSubmittable': True
+        }
+    )
+    def assembly(self, request, reference_files):
+        assembly_set = set()
+        assembly = ''
+        for ref_file in reference_files:
+            ref_file_object = request.embed(ref_file, '@@object?skip_calculated=true')
+            if ref_file_object['content_type'] == 'genome reference':
```
zhwshen, line 717: Not sure if genome index will make sense to add here because single cell pipeline uses genome index, but for tabular file output. Matrix file is only RNA.

shengchd: Added, seems no hurt to add an OR case

src/igvfd/types/file.py
```diff
@@ -680,6 +680,48 @@ def summary(self, request, content_summary, file_set, filtered=None, analysis_st
...
+            'notSubmittable': True
+        }
+    )
+    def assembly(self, request, reference_files):
+        assembly_set = set()
+        assembly = ''
+        for ref_file in reference_files:
+            ref_file_object = request.embed(ref_file, '@@object?skip_calculated=true')
+            if ref_file_object['content_type'] == 'genome reference' or ref_file_object['content_type'] == 'genome index':
+                assembly_set.add(ref_file_object.get('assembly', ''))
```
jenjou, line 718: Same comment as above about the transcriptome annotation set.

src/igvfd/types/file.py
```diff
@@ -680,6 +680,48 @@ def summary(self, request, content_summary, file_set, filtered=None, analysis_st
...
+            'notSubmittable': True
+        }
+    )
+    def transcriptome_annotation(self, request, reference_files):
+        transcriptome_annotation_set = set()
+        transcriptome_annotation = ''
+        for ref_file in reference_files:
+            ref_file_object = request.embed(ref_file, '@@object?skip_calculated=true')
+            if ref_file_object['content_type'] == 'transcriptome reference' or ref_file_object['content_type'] == 'transcriptome index':
+                transcriptome_annotation_set.add(ref_file_object.get('transcriptome_annotation', ''))
```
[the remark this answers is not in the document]
shengchd: yeah I thought it would be good to capture it as a mixed case, where one transcriptome ref file is missing `transcriptome_annotation` and the other has a value in transcriptome_annotation, if that makes sense.

PR #1610, opened 2025-06-27
IGVF-2837-analysis-targeted-genes

ian-whaling:


src/igvfd/types/file_set.py
```diff
@@ -965,6 +966,31 @@ def workflows(self, request, files=None):
...
+                'title': 'Targeted Gene',
+                'type': 'string',
+                'linkTo': 'Gene'
+            }
+        }
+    )
+    def targeted_genes(self, request, input_file_sets=None):
+        analysis_set_targeted_genes = set()
+        for input_file_set in input_file_sets:
+            if input_file_set.startswith('/measurement-sets/'):
```
shengchd, line 988: are we only considering measurement sets here not including e.g. intermediate analysis sets for now?

shengchd (commented):
Might need a safe guard for None case

src/igvfd/types/file_set.py
```diff
@@ -965,6 +966,31 @@ def workflows(self, request, files=None):
...
+            'items': {
+                'title': 'Targeted Gene',
+                'type': 'string',
+                'linkTo': 'Gene'
+            }
+        }
+    )
+    def targeted_genes(self, request, input_file_sets=None):
+        analysis_set_targeted_genes = set()
+        for input_file_set in input_file_sets:
```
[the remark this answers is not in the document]
ian-whaling: fixed.

src/igvfd/types/file_set.py
ian-whaling: you're right it should also account for that.

shengchd (approved):

PR #1618, opened 2025-07-07
IGVF-2813-add-immune-sge-label

shengchd:


jenjou (dismissed):
Code looks good, but I think this will be conflicting with IGVF-2820.

shengchd:
> Code looks good, but I think this will be conflicting with IGVF-2820.

thanks I can rebase on IGVF-2820.

jenjou (approved):

PR #1625, opened 2025-07-08
IGVF-2813-add-immune-sge-label-2

shengchd:


src/igvfd/tests/data/inserts/assay_term.json
```diff
@@ -45,6 +45,10 @@
         "uuid": "0393b224-8f87-5e94-7aeb-faff044f707d",
         "aliases": [
             "igvf:CRISPR"
+        ],
+        "preferred_assay_titles": [
```
keenangraham, line 49: Do these need to be `preferred_assay_title`?

shengchd: we are change preferred_assay_title -> preferred_assay_titles under measurement set

PR #1634, opened 2025-07-12
IGVF-2865-aset-summary-bug

jenjou:


src/igvfd/types/file_set.py
```diff
@@ -905,7 +906,7 @@ def sample_summary(self, request, samples=None):
         additional_phrase_suffix = ''
         if additional_phrases_joined:
             additional_phrase_suffix = f', {additional_phrases_joined}'
-        summary = f"{', '.join(taxa)} {', '.join(all_sample_terms)}{additional_phrase_suffix}"
+        summary = f"{taxa_phrase} {', '.join(all_sample_terms)}{additional_phrase_suffix}"
```
shengchd, line 909: this would be an edge/invalid case, but if taxa_phrase is an empty string, it would have an extra space at the beginning?

jenjou: I can move the space into the taxa phrase itself.

shengchd (approved):

PR #1638, opened 2025-07-14
IGVF-2851-anaset-pipe-stat

zhwshen:


src/igvfd/schemas/analysis_set.json
```diff
@@ -114,6 +114,28 @@
...
+                "preprocessing",
+                "processing",
+                "completed",
+                "not applicable"
+            ],
+            "enum_descriptions": {
+                "preprocessing": "The analysis set has been created with `input_file_sets` and is awaiting processing.",
+                "processing": "The analysis set is currently being processed by the single cell uniform pipeline on Terra.bio.",
+                "completed": "The analysis set has been successfully processed by the pipeline.",
+                "not applicable": "The analysis set is not applicable for single cell uniform pipeline processing, e.g., it is an MPRA uniform pipeline analysis set."
```
ian-whaling, line 133: Why even have this option?

src/igvfd/schemas/analysis_set.json
```diff
@@ -114,6 +114,28 @@
...
+            "permission": "admin_only",
+            "enum": [
+                "preprocessing",
+                "processing",
+                "completed",
+                "not applicable"
+            ],
+            "enum_descriptions": {
+                "preprocessing": "The analysis set has been created with `input_file_sets` and is awaiting processing.",
+                "processing": "The analysis set is currently being processed by the single cell uniform pipeline on Terra.bio.",
```
ian-whaling, line 131: Do we want to allow this for the Perturb-seq pipeline as well? They have to pre-define their analysis sets for processing. (even if we are not doing the processing)

zhwshen: Technically, this property is for managing single cell data setup, running, and cleanup. I would like to have way of telling if an AnalysisSet that is uniformly processed directly. Using this pipeline status is an option (AnalysisSet.pipeline_status == completed >> AnalysisSet.files.upload_status >> if validated, delete from Terra). Otherwise, I will need to go AnalysisSet >> if uniform_pipeline==True >> input_file_sets.file_set.assay_term >> if single cell == True >> AnalysisSet.files.upload_status >> if validated, delete from Terra.

src/igvfd/schemas/analysis_set.json
```diff
@@ -114,6 +114,27 @@
                 "appscript": "[\"IGVFSM02881437\", \"john-doe:tissue01\"]",
                 "igvf_utils": "IGVFSM02881437 or john-doe:tissue01"
             }
+        },
+        "uniform_pipeline_status": {
+            "title": "Uniform Pipeline Status",
+            "description": "The status of the single cell or PERTURB-seq uniform pipeline processing for this analysis set, if applicable.",
```
ian-whaling, line 120: Perturb-seq not PERTURB-seq

ian-whaling (changes requested):

ian-whaling (dismissed):

PR #1641, opened 2025-07-17
IGVF-2885-rm-assembly-align-signal

jenjou:


shengchd (commented):
code looks good, just have a comment on dependency test code. Might also make sense to move the dependency out of file.json, to shared_dependencies.json? (can be done later)

shengchd (approved):

PR #1664, opened 2025-07-31
IGVF-2897-calc-assembly-align-signal

jenjou:


src/igvfd/types/file.py
```diff
@@ -1188,7 +1340,8 @@ def assembly(self, request, derived_from):
         }
     )
     def transcriptome_annotation(self, request, derived_from):
-        parent_file_object = request.embed(derived_from[0], '@@object?skip_calculated=true')
+        parent_file_object = request.embed(
+            derived_from[0], '@@object_with_select_calculated_properties?field=transcriptome_annotation')
```
shengchd, line 1344: probably want to do the same thing for alignment and signal file summaries, add assembly and transcriptome back?

shengchd: they were removed from here: https://github.com/IGVF-DACC/igvfd/pull/1641/files#diff-01e44507c3148abef1d1d7b8ffc39d8f8296c41e1d601cca4fadebf8392de766

src/igvfd/types/file.py
```diff
@@ -644,6 +644,84 @@ def summary(self, request, content_summary, file_set, assembly=None, transcripto
              if x is not None]
         )
 
+    @calculated_property(
+        define=True,
```
shengchd, line 648: can you explain to me how define=True is working here?

jenjou: Based on what I understand, including define=True will make sure this property is calculated in advance, therefore it could be used by the other calculated properties. Another place where it is used is the `content_summary` calculated property on Matrix File.

shengchd: got it thanks

PR #1667, opened 2025-08-01
IGVF-2918-parse-perturb-seq

ian-whaling:


src/igvfd/schemas/changelogs/assay_term.md
```diff
@@ -1,5 +1,10 @@
 ## Changelog for *`assay_term.json`*
 
+### Minor changes since schema version 17
```
jenjou, line 3: Minor issue - should be version 18 not 17

jenjou (dismissed):

PR #1689, opened 2025-08-19
IGVF-2755-file-batch-download

keenangraham:


src/igvfd/tests/test_metadata_metadata.py
```diff
@@ -691,6 +806,119 @@ def test_metadata_metadata_report_split_column_and_fields_by_experiment_and_file
...
+            ('File format', ['file_format']),
+            ('File format type', ['file_format_type']),
+            ('File content type', ['content_type']),
+            ('File summary', ['summary']),
+            ('Cell type annotation', ['cell_type_annotation.term_name']),
+            ('Creation timestamp', ['creation_timestamp']),
+            ('File size', ['file_size']),
+            ('File S3 URI', ['s3_uri']),
+            ('File assembly', ['assembly']),
+            ('File transcriptome annotation', ['transcritome_annotation']),
```
ottojolanki, line 830: typo also here, so not catching the above typo.

src/igvfd/metadata/constants.py
```diff
@@ -71,3 +86,114 @@
...
+        ('Samples', ['samples.accession']),
+        ('Sample term names', ['samples.sample_terms.term_name']),
+        ('Sample summaries', ['samples.summary']),
+        ('Cell type annotation', ['files.cell_type_annotation.term_name']),
+        ('Creation timestamp', ['files.creation_timestamp']),
+        ('File size', ['files.file_size']),
+        ('Fileset lab', ['lab.title']),
+        ('File S3 URI', ['files.s3_uri']),
+        ('File assembly', ['files.assembly']),
+        ('File transcriptome annotation', ['files.transcritome_annotation']),
```
ottojolanki, line 114: typo, also in tests so not caught there either.

src/igvfd/metadata/decorators.py
```diff
@@ -20,3 +20,19 @@ def wrapper(context, request):
...
         return wrapper
     return decorator
+
+
+def allowed_types_v2(types):
+    def decorator(func):
+        @wraps(func)
+        def wrapper(context, request):
+            qs = QueryString(request)
+            type_filters = qs.get_type_filters()
```
ottojolanki, line 30: Do we need to check that `len(type_filters) > 0` here?

ottojolanki (changes requested):
Fix typo. One question about allowing empty type_filter.

src/igvfd/tests/test_metadata_metadata.py
keenangraham: Nice, fixed.

src/igvfd/metadata/decorators.py
keenangraham: Should return empty list.

PR #1696, opened 2025-08-22
IGVF-2963-validate-onlist

ian-whaling:


src/igvfd/types/file.py
```diff
@@ -1107,24 +1107,44 @@ def summary(self, request, file_set, content_type, seqspec_of=None, analysis_ste
         }
     )
     def validate_onlist_files(self, request, content_type, seqspec_of=None):
-        # Validate onlist files if the file is not linked to a single cell assay measurement set
+        # Validate onlist files if the file is linked to a single cell assay measurement set
+
+        single_cell_assay_term_names = list(SINGLE_CELL_ASSAY_TERMS.values()) + [
```
zhwshen, line 1112: Should it stay with using assay term IDs which is less likely to change?

src/igvfd/tests/test_types_configuration_file.py
```diff
@@ -0,0 +1,49 @@
...
+    res = testapp.get(configuration_file_seqspec['@id'])
+    assert res.json.get('validate_onlist_files', '') is False
+    testapp.patch_json(
+        measurement_set_perturb_seq['@id'],
+        {
+            'auxiliary_sets': [base_auxiliary_set['@id']]
+        }
+    )
+    res = testapp.get(configuration_file_seqspec['@id'])
+    assert res.json.get('validate_onlist_files', '') is True
```
zhwshen, line 49: May consider adding a scenario for a non-single cell and non-perturb-seq measurement or aux set (e.g., MPRA) to make sure it calculates right.

ian-whaling: added more tests

PR #1734, opened 2025-09-22
IGVF-3023-t2t-transcript-ann-enum

jenjou:


src/igvfd/schemas/file.json
```diff
@@ -286,6 +286,48 @@
...
+                                ]
+                            }
+                        }
+                    }
+                },
+                {
+                    "if": {
+                        "properties": {
+                            "assembly": {
+                                "const": "C57BL_6J_T2T_v1"
```
shengchd, line 315: this is C57BL_6J_T2T_v1 + GRCm39_X actually?

PR #1736, opened 2025-09-23
IGVF-3018-remodel-stiffness

ian-whaling:


jenjou (changes requested):
Just one change - schema version in Treatment needs to be bumped.

jenjou (approved):

PR #1756, opened 2025-10-02
IGVF-3027-methyl-reads-no-ref

jenjou:


src/igvfd/types/file.py
```diff
@@ -739,7 +739,7 @@ def summary(self, request, content_summary, file_set, assembly=None, transcripto
             'notSubmittable': True
         }
     )
-    def transcriptome_annotation(self, request, reference_files):
+    def transcriptome_annotation(self, request, reference_files=[]):
```
shengchd, line 742: think we prefer to using None as a default value for input parameter?

src/igvfd/types/file.py
```diff
@@ -761,7 +761,7 @@ def transcriptome_annotation(self, request, reference_files):
             'notSubmittable': True
         }
     )
-    def assembly(self, request, reference_files):
+    def assembly(self, request, reference_files=[]):
```
shengchd, line 764: same here

src/igvfd/schemas/alignment_file.json
```diff
@@ -75,6 +74,39 @@
...
+                            "enum": [
+                                "alignments",
+                                "alignments with modifications",
+                                "transcriptome alignments"
+                            ]
+                        }
+                    }
+                },
+                {
+                    "not": {
```
shengchd, line 96: is this dependency saying if content_type is methylated reads, you can't specify reference_files? which is more strict than not required but it's fine?

shengchd (commented):
just have a comment on default value for input parameter

shengchd (approved):

PR #1987, opened 2026-03-27
IGVF-3412-deliveries

ian-whaling:


src/igvfd/searches/configs/biosample.py
```diff
@@ -69,8 +69,8 @@ def biosample():
             'construct_library_sets.associated_phenotypes.term_name': {
                 'title': 'Associated Phenotype'
             },
-            'nucleic_acid_delivery': {
-                'title': 'Nucleic Acid Delivery Method'
+            'construct_delivery_methods': {
+                'title': 'Construct Delivery Method'
```
zhwshen: Method**s** plural? Same for all the other search configs.

src/igvfd/schemas/changelogs/biosample.md
```diff
@@ -1,5 +1,6 @@
 ## Changelog for *`biosample.json`*
 
+* Rename `nucleic_acid_delivery` to `construct_delivery_methods`. (04/08/2026)
```
zhwshen: Need to change the time to 4-15.

src/igvfd/schemas/changelogs/sample.md
```diff
@@ -1,5 +1,6 @@
 ## Changelog for *`sample.json`*
 
+* Rename `nucleic_acid_delivery` to `construct_delivery_methods`. (04/08/2026)
```
zhwshen: Same with the date change as above.

src/igvfd/schemas/multiplexed_sample.json
```diff
@@ -153,8 +153,8 @@
             "comment": "Do not submit. MOI should only be specified on individual samples contributing to a multiplexed sample object.",
             "permission": "admin_only"
         },
-        "nucleic_acid_delivery": {
-            "comment": "Do not submit. Nucleic acid delivery should only be specified on individual samples contributing to a multiplexed sample object.",
+        "construct_delivery_methods": {
```
zhwshen: Is this needed if this is calculated from the subsamples? Or is it because it is a dependency?

src/igvfd/upgrade/sample.py
```diff
@@ -600,3 +600,21 @@ def tissue_23_24(value, system):
         notes += f' This sample had the ccf_id {value.get("ccf_id", "")}, which has been removed via upgrade.'
         value['notes'] = notes.strip()
         del value['ccf_id']
+
+
+@upgrade_step('primary_cell', '24', '25')
```
zhwshen, line 615: Does it matter if the upgrade is sample 24 to 25 (primary cell and tissue), but it is a different version for other sample obj types? If this is an acceptable format, then it is OK.

zhwshen (changes requested):

src/igvfd/schemas/multiplexed_sample.json
ian-whaling: It is inherited from sample

src/igvfd/upgrade/sample.py
ian-whaling: The name of the function doesn't really matter, so long as it is unique and interpretable. There is no convention in place when there are mixed versions

src/igvfd/searches/configs/biosample.py
ian-whaling: The name for facets is singular regardless of the property being plural.

src/igvfd/schemas/changelogs/sample.md
ian-whaling: fixed

src/igvfd/schemas/changelogs/biosample.md
ian-whaling: fixed

src/igvfd/searches/configs/biosample.py
zhwshen: It should be plural. Checked with @khzlin on another ticket that adds an array property to search config.

ian-whaling: pluralized it

zhwshen (dismissed):

PR #1989, opened 2026-03-30
IGVF-3391-curated-set-assay-term

jenjou:


shengchd (approved):
code looks good, I'm wondering if we need a future ticket for audits on if preferred_assay_titles matching with assay_term here.
Actually missing changelogs.

shengchd (changes requested):
missing changelogs

shengchd (dismissed):

src/igvfd/schemas/curated_set.json
```diff
@@ -55,7 +58,40 @@
...
+        ],
+        "file_set_type": {
+            "if": {
+                "properties": {
+                    "file_set_type": {
+                        "const": "external sequencing data"
+                    }
+                }
+            },
+            "then": {
```
shengchd: I thought preferred_assay_titles/assay_term should be optional, but not required when file_set_type = 'external sequencing data'?

shengchd (dismissed):

shengchd (approved):

PR #1998, opened 2026-04-02
IGVF-2887-tab-file-calc-assembly

jenjou:


src/igvfd/types/file.py
```diff
@@ -1255,6 +1255,52 @@ def enrichment_design_for(self, request, enrichment_design_for):
...
+    def transcriptome_annotation(self, request, reference_files=None):
+        transcriptome_annotation_set = set()
+        transcriptome_annotation = None
+        if reference_files is not None:
+            for ref_file in paths_filtered_by_status(request, reference_files):
+                ref_file_object = request.embed(ref_file, '@@object?skip_calculated=true')
+                if ref_file_object['content_type'] == 'transcriptome reference' or ref_file_object['content_type'] == 'transcriptome index':
+                    transcriptome_annotation_set.add(ref_file_object.get('transcriptome_annotation', None))
+        if len(transcriptome_annotation_set) > 1:
+            transcriptome_annotation = 'Mixed transcriptome annotations'
```
zhwshen, line 1281: Does it make sense to list all or use mixed? I assume that annotation total per file shouldn't be more than 1 or 2 at best?

src/igvfd/types/file.py
```diff
@@ -1255,6 +1255,52 @@ def enrichment_design_for(self, request, enrichment_design_for):
...
+    )
+    def assembly(self, request, reference_files=None):
+        assembly_set = set()
+        assembly = None
+        if reference_files is not None:
+            for ref_file in paths_filtered_by_status(request, reference_files):
+                ref_file_object = request.embed(ref_file, '@@object?skip_calculated=true')
+                if ref_file_object['content_type'] == 'genome reference' or ref_file_object['content_type'] == 'genome index':
+                    assembly_set.add(ref_file_object.get('assembly', None))
+        if len(assembly_set) > 1:
```
zhwshen, line 1303: Same as above. Probably like mm10 + GRCm39 or GRCh38+mm10 in the mixed species case?

src/igvfd/upgrade/file.py
```diff
@@ -680,3 +680,96 @@ def reference_file_22_23(value, system):
...
+        notes += f' The submitted assembly {value.get("assembly", "")} was moved to the submitted_assembly property.'
+        del value['assembly']
+    if 'transcriptome_annotation' in value:
+        value['submitted_transcriptome_annotation'] = value.get('transcriptome_annotation', '')
+        notes += f' The submitted transcriptome annotation {value.get("transcriptome_annotation", "")} was moved to the submitted_transcriptome_annotation property.'
+        del value['transcriptome_annotation']
+    if 'reference_files' in value or value['content_type'] in excluded_content_types:
+        return
+    else:
+        if value['status'] == 'deleted':
```
zhwshen: Why is revoked not done the same way as deleted?

src/igvfd/upgrade/file.py
```diff
@@ -680,3 +680,96 @@ def reference_file_22_23(value, system):
 
     if notes.strip() != '':
         value['notes'] = notes.strip()
+
+
+@upgrade_step('tabular_file', '20', '21')
+def tabular_file_20_21(value, system):
+    # https://igvf.atlassian.net/browse/IGVF-2887
+    excluded_content_types = [
```
zhwshen: I wonder if this file list should be a constant and then get imported here, so any future reference to this list can be done in one place.

src/igvfd/tests/test_types_file.py
```diff
@@ -249,6 +269,26 @@ def test_types_file_calculated_assembly(testapp, matrix_file, signal_file, align
...
+    res = testapp.get(tabular_file['@id'])
+    assert res.json.get('assembly') == 'GRCh38'
+    testapp.patch_json(
+        tabular_file['@id'],
+        {
+            'reference_files': [reference_file['@id'], reference_file_with_assembly['@id']]
+        }
+    )
+    res = testapp.get(tabular_file['@id'])
+    assert res.json.get('assembly') == 'Mixed genome assemblies'
```
zhwshen, line 290: This should be changed only if ending up listing all assemblies.

zhwshen (changes requested):

src/igvfd/types/file.py
jenjou: I have tried to keep the code as similar to the other cases of calculated `transcriptome_annotation` as possible. Here is one example for Alignment File: https://github.com/IGVF-DACC/igvfd/blob/dev/src/igvfd/types/file.py#L762-L773

So if we change it, I would prefer to change how it works for all other schemas too.

src/igvfd/upgrade/file.py
jenjou: I guess the deleted vs not deleted case are actually handled the same as you pointed out, so I will just merge into a single case here and will not check the file's status.

src/igvfd/upgrade/file.py
jenjou: I'm unsure how to do this, would have to ask Keenan for guidance, so I'll try it again on Monday.

keenangraham: One issue with doing this is the list could change in the future, which would modify the upgrade function that already ran. I think in this case you might just want to capture the current state of the system and keep the list defined locally.

zhwshen: @keenangraham @jenjou Does it make sense to rename this variable as `excluded_content_type_v21` and make a note? In case future lists might change?

keenangraham: That would be one option if you did need to share it (maybe with a comment that says do not modify?). If it's just local to the function the name is probably fine?

zhwshen (dismissed):

src/igvfd/audit/file.py
```diff
@@ -158,7 +158,17 @@ def audit_file_mixed_assembly_transcriptome_annotation(value, system):
             'GENCODE 47'
         ],
         'GRCh38, mm10': [
-            'GENCODE 32, GENCODE M23'
+            'GENCODE 32, GENCODE M23',
+            'GENCODE 41, GENCODE M25'
+        ],
+        '"GRCh38, GRCm39': [
```
keenangraham: Is there supposed to be a double quotes here? `"`

zhwshen (dismissed):

PR #2001, opened 2026-04-06
IGVF-3415-add-file-content-type

khzlin:


shengchd (changes requested):
please also update the AS changelogs

shengchd (approved):

PR #2005, opened 2026-04-08
IGVF-3419-software-regex

zhwshen:


shengchd (commented):
Do we need an upgrade for this since the regex is more strict now? will also need changelog updates

shengchd (dismissed):

shengchd (approved):

PR #2008, opened 2026-04-10
IGVF-3430-in-vivo

ian-whaling:
…uded


jenjou (commented):
Code looks OK. Can you add a test insert for the new enums?

jenjou (dismissed):

PR #2010, opened 2026-04-13
IGVF-3399-pseudobulk-UI

khzlin:


jenjou (approved):
Looks fine to me. Coveralls is failing, but the sections it points to are not the files modified in this PR, so I don't think it is relevant.

PR #2014, opened 2026-04-15
IGVF-3443-software-regex-bug

zhwshen:


keenangraham:
Should we just add `A-Z` to the regex instead of converting to lowercase?

src/igvfd/schemas/software.json
```diff
@@ -54,7 +54,7 @@
             "title": "Name",
             "description": "Unique name of the software package; a lowercase version of the title.",
```
ottojolanki: Do we want to change this, or do we still want to encourage using lowercase?

ottojolanki: With the new regex this is not accurate and might cause some confusion down the road.

ottojolanki (approved):

PR #2020, opened 2026-04-16
IGVF-3405-simple-samples

zhwshen:


src/igvfd/types/file_set.py
```diff
@@ -874,204 +1147,141 @@ def protocols(self, request, input_file_sets=None):
...
-            'cell line, pooled cell specimen': ['pooled cell specimen']
-        }
-
-        classification_to_prefix = {
-            'differentiated cell specimen': 'differentiated',
-            'reprogrammed cell specimen': 'reprogrammed',
-            'pooled differentiated cell specimen': 'pooled differentiated',
-            'pooled reprogrammed cell specimen': 'pooled reprogrammed'
+        if not samples:
+            return ''
```
jenjou: Should this be `''` or should it be `None`?

I think the former would result in the calculated property existing in every object but it will look like: `"sample_summary": ""` rather than the property not appearing at all. Not sure what was intended here.

src/igvfd/types/file_set.py
```diff
@@ -874,204 +1147,141 @@ def protocols(self, request, input_file_sets=None):
...
-                if 'induced to' in terms_by_classification:
-                    terms_by_classification = terms_by_classification.replace(
-                        'induced to', f'{classification} induced to'
+        sample_objects = [request.embed(sample, '@@object') for sample in samples]
+        all_classifications = set()
+        taxa_values = []
+        for sample_object in sample_objects:
+            all_classifications.update(sample_object.get('classifications', []))
+            sample_taxa = sample_object.get('taxa', [])
+            if isinstance(sample_taxa, (list, tuple, set)):
```
jenjou: I think taxa is a string type on all Samples (calculated for everything except Technical Sample where it is submitted). Could you clarify when the list/tuple/set case appears?

src/igvfd/types/file_set.py
```diff
@@ -874,204 +1147,141 @@ def protocols(self, request, input_file_sets=None):
...
-                # Insert the classification before the targeted_sample_term if it exists.
-                if 'induced to' in terms_by_classification:
-                    terms_by_classification = terms_by_classification.replace(
-                        'induced to', f'{classification} induced to'
+        sample_objects = [request.embed(sample, '@@object') for sample in samples]
+        all_classifications = set()
+        taxa_values = []
+        for sample_object in sample_objects:
+            all_classifications.update(sample_object.get('classifications', []))
+            sample_taxa = sample_object.get('taxa', [])
```
jenjou: The default return value should match the type of the property, in this case I think it makes sense to have `sample_object.get('taxa', '')`.

src/igvfd/types/file_set.py
```diff
@@ -874,204 +1147,141 @@ def protocols(self, request, input_file_sets=None):
...
-        classification_to_prefix = {
-            'differentiated cell specimen': 'differentiated',
-            'reprogrammed cell specimen': 'reprogrammed',
-            'pooled differentiated cell specimen': 'pooled differentiated',
-            'pooled reprogrammed cell specimen': 'pooled reprogrammed'
+        if not samples:
+            return ''
+
+        disease_term_prefixes = ('DOID:', 'EFO:', 'HP:', 'MONDO:')
+        taxa_to_label = {
```
jenjou: Since this dictionary is only utilized by the _sample_summary_normalize_taxa() function, would it make sense to move it within that function instead of having it as an argument each time the function is called?

src/igvfd/types/file_set.py
```diff
@@ -110,6 +114,275 @@ def get_file_set_props_for_summary_and_samples(request, file_set):
...
+        'strains': donor_strains,
+        'taxa': donor_taxa,
+    }
+
+
+def _sample_summary_normalize_taxa(taxa_values, taxa_to_label):
+    if not taxa_values:
+        return ''
+
+    if not isinstance(taxa_values, (list, tuple, set)):
```
jenjou: Similar to my other comment, just wanted to double check where the "list" type taxa appears?

src/igvfd/types/file_set.py
zhwshen: You are right. I was thinking of the mixed species cases and thought they would be a list. Double-checked, it is just `mixed species` as an enum.

src/igvfd/types/file_set.py
```diff
@@ -110,6 +114,275 @@ def get_file_set_props_for_summary_and_samples(request, file_set):
...
+        phrase = f'{disease_prefix}{sample_term_phrase} tissue'.strip()
+    else:
+        organoid_type = get_first_matching_classification(
+            sample_classifications,
+            ['gastruloid', 'embryoid', 'organoid']
+        )
+        if organoid_type:
+            phrase = f'{organoid_sample_term} {organoid_type}'.strip()
+        elif 'primary cell' in sample_classifications:
+            if targeted_sample_term_phrase:
```
jenjou: I think this case is impossible for Primary Cell, since they do not have the `targeted_sample_term` property. So this section can probably be simplified as they'll never have any targeted sample term phrase.

src/igvfd/types/file_set.py
```diff
@@ -874,204 +1147,141 @@ def protocols(self, request, input_file_sets=None):
...
+            donor_data = _sample_summary_get_donor_data(request, sample_object)
+            donor_accessions.update(donor_data['accessions'])
+            donor_strains.update(donor_data['strains'])
+            donor_taxa.update(donor_data['taxa'])
+
+        donor_phrase = ''
+        is_all_mouse_donors = donor_taxa and donor_taxa == {'Mus musculus'}
+        if is_all_mouse_donors and donor_strains:
+            if len(donor_strains) == 1:
+                donor_phrase = f' from {sorted(donor_strains)[0]} mouse'
```
jenjou: Could drop the final word "mouse" here, since the taxa is at the beginning of the summary, it results in a bit of redundancy: "Mouse brain tissue from C57BL/6 mouse"

src/igvfd/types/file_set.py
```diff
@@ -110,6 +114,275 @@ def get_file_set_props_for_summary_and_samples(request, file_set):
...
+        return system_slims[0]
+    organ_slims = term_obj.get('organ_slims', [])
+    if organ_slims:
+        return organ_slims[0]
+    return ''
+
+
+def _sample_summary_build_tissue_group_phrase(request, tissue_sample_objects, disease_prefixes):
+    def _format_tissue_phrase(term_name):
+        if term_name.endswith(' tissue') or term_name.endswith(' tissues'):
```
jenjou: Minor comment: endswith() can accept a tuple so this could be: `term_name.endswith((' tissue', 'tissues'))`

src/igvfd/types/file_set.py
zhwshen: I thought it might make sense to go with the sample_summary function in case the helper func might get used somewhere else. Probably doesn't hurt to move it up.

src/igvfd/types/file_set.py
```diff
@@ -110,6 +114,275 @@ def get_file_set_props_for_summary_and_samples(request, file_set):
...
+            return term_name
+        return f'{term_name} tissue'
+
+    all_term_names = set()
+    for sample_obj in tissue_sample_objects:
+        for term_path in sample_obj.get('sample_terms', []):
+            term_name = _sample_summary_get_term_name(request, term_path)
+            if term_name:
+                all_term_names.add(term_name)
```
jenjou, line 267: This part seems similar to what is in  _sample_summary_get_sample_term_names() , I'm having trouble understanding why the "singular" version _sample_summary_get_term_name() is used here.

src/igvfd/types/file_set.py
```diff
@@ -110,6 +114,275 @@ def get_file_set_props_for_summary_and_samples(request, file_set):
...
+
+
+def _sample_summary_get_slim_for_sample_term(request, term_path):
+    term_obj = request.embed(
+        term_path,
+        '@@object_with_select_calculated_properties?field=system_slims&field=organ_slims'
+    )
+    system_slims = term_obj.get('system_slims', [])
+    if system_slims:
+        return system_slims[0]
```
jenjou, line 248: Is it always guaranteed that when we request system_slims it is always returned in the same order? I am just wondering if we might repeat this request and the first time "digestive system" is first but the 2nd time, "endocrine system" is first in the array. That could cause two samples to end up grouped in different slim categories.

src/igvfd/types/file_set.py
```diff
@@ -110,6 +114,275 @@ def get_file_set_props_for_summary_and_samples(request, file_set):
...
+                break
+        if slim:
+            slim_groups[slim] = slim_groups.get(slim, 0) + 1
+        else:
+            for term_path in sample_obj.get('sample_terms', []):
+                term_name = _sample_summary_get_term_name(request, term_path)
+                if term_name:
+                    no_slim_term_names.add(term_name)
+    parts = []
+    for slim in sorted(slim_groups):
```
jenjou: Here, since `parts` is going to add another set of terms from the "no slim term names", I think you can skip the sorted() here and add sorted() at the end, in line 327.

src/igvfd/types/file_set.py
```diff
@@ -110,6 +114,275 @@ def get_file_set_props_for_summary_and_samples(request, file_set):
...
+                term
+                for disease_terms in samples_disease_terms
+                for term in disease_terms
+            ))
+            disease_phrase = _sample_summary_format_disease_phrase(group_disease_terms)
+            if disease_phrase:
+                return f'{disease_phrase} {tissue_phrase}'.strip()
+        return tissue_phrase
+
+    # For >2 unique tissue terms, use disease-based grouping
```
jenjou: Need to revisit this part. I think if there is a mix of samples (e.g. 2 have disease, 1 has nothing), this means the summary only returns a phrase mentioning the 2 with disease and the "nothing" sample is excluded.

src/igvfd/types/file_set.py
zhwshen: _sample_summary_get_term_name just takes an ontology term and does the query. _sample_summary_get_sample_term_names specifically looks for `sample.sample_terms`. Another function _sample_summary_get_targeted_sample_term_name does `sample.target_sample_term`. It is basically using a query func as a helper func. I can just do the query step in each function instead of calling this.

PR #2021, opened 2026-04-16
IGVF-3431-preferred-assay-slims

jenjou:


src/igvfd/types/file_set.py
```diff
@@ -711,6 +795,29 @@ def preferred_assay_titles(self, request, input_file_sets=None):
...
 
+    @calculated_property(
+        schema={
+            'title': 'Preferred Assay Slims',
+            'description': 'Preferred Assay Slim(s) of assays that produced data analyzed in the analysis set.',
+            'type': 'array',
+            'minItems': 1,
+            'uniqueItems': True,
+            'items': {
+                'title': 'Preferred Assay Slims',
```
zhwshen: should this be `Preferred Assay Slim` or the plural form of it?

src/igvfd/types/file_set.py
```diff
@@ -110,6 +110,90 @@ def get_file_set_props_for_summary_and_samples(request, file_set):
...
+        'ONT Fiber-seq': ['gene expression', 'long read'],
+        'ONT direct WGS': ['gene expression', 'long read'],
+        'Bisulfite-seq': ['DNA methylation'],
+        'WGS': ['genetic profiling'],
+        'DNase-seq': ['chromatin accessibility']
+    }
+    if title in title_to_slims:
+        return title_to_slims[title]
+    else:
+        return []
```
zhwshen: Are we expecting cases where there is no `preferred_assay_slims`? If so, should it be an empty list or something like `Miscellaneous`?

src/igvfd/types/file_set.py
```diff
@@ -711,6 +795,29 @@ def preferred_assay_titles(self, request, input_file_sets=None):
...
+            'uniqueItems': True,
+            'items': {
+                'title': 'Preferred Assay Slims',
+                'description': 'Category of assay that produced data analyzed in the analysis set.',
+                'type': 'string'
+            },
+            'notSubmittable': True,
+        }
+    )
+    def preferred_assay_slims(self, request, preferred_assay_titles=None):
```
zhwshen, line 820: Consider writing the whole function into a helper function, so it is not repeated multiple times in the same .py file for different file set types.

src/igvfd/types/file.py
```diff
@@ -380,6 +380,27 @@ def preferred_assay_titles(self, request, file_set):
...
 
+    @calculated_property(
+        schema={
+            'title': 'Preferred Assay Slims',
+            'description': 'Preferred assay slims from the file set this file belongs to.',
+            'type': 'array',
+            'minItems': 1,
+            'uniqueItems': True,
+            'items': {
+                'title': 'Assay Slim',
```
zhwshen: Should this be Preferred Assay Slim or Assay Slim? There is an assay slim as well.

zhwshen (changes requested):

src/igvfd/types/file_set.py
jenjou: There probably shouldn't be if we are updating the code as we add new preferred assay titles. I just thought to include it in case we miss it, somehow.

Do we need there to always be at least 1 slim?

jenjou: Ok, confirmed with Khine to add some default value.

src/igvfd/types/file_set.py
jenjou: Will fix typo

src/igvfd/types/file.py
jenjou: Renamed it.

zhwshen:
@jenjou One more question, should this go into the base class of File and FileSet as calculated?

zhwshen (dismissed):

zhwshen (dismissed):

PR #2022, opened 2026-04-16
IGVF-3421-assay-term-model-set

shengchd:


src/igvfd/mappings/analysis_set.json
```diff
@@ -328,6 +328,9 @@
                             "assay_term": {
                                 "type": "keyword"
                             },
+                            "assay_terms": {
```
zhwshen, line 331: This seems confusing to me that it has both an assay_term and an assay_terms.

src/igvfd/mappings/auxiliary_set.json
```diff
@@ -327,6 +327,9 @@
                             "assay_term": {
                                 "type": "keyword"
                             },
+                            "assay_terms": {
```
zhwshen, line 330: Same as above.

src/igvfd/mappings/construct_library_set.json
```diff
@@ -355,6 +355,9 @@
                             "assay_term": {
                                 "type": "keyword"
                             },
+                            "assay_terms": {
```
zhwshen, line 358: Same as above.

src/igvfd/mappings/auxiliary_set.json
```diff
@@ -2195,6 +2201,9 @@
...
```
zhwshen, line 2216: Same as above.

src/igvfd/mappings/measurement_set.json
```diff
@@ -558,6 +558,9 @@
                             "assay_term": {
                                 "type": "keyword"
                             },
+                            "assay_terms": {
```
zhwshen, line 561: Same as above.

src/igvfd/mappings/measurement_set.json
```diff
@@ -1324,6 +1327,9 @@
...
```
zhwshen, line 1336: Same as above.

src/igvfd/mappings/measurement_set.json
```diff
@@ -1693,6 +1699,9 @@
...
+                                "type": "keyword"
```
zhwshen, line 1709: Same as above.

src/igvfd/mappings/measurement_set.json
```diff
@@ -2882,6 +2891,9 @@
                             "assay_term": {
                                 "type": "keyword"
                             },
+                            "assay_terms": {
```
zhwshen, line 2906: Same as above.

src/igvfd/mappings/prediction_set.json
```diff
@@ -506,6 +506,9 @@
                             "assay_term": {
                                 "type": "keyword"
                             },
+                            "assay_terms": {
```
zhwshen, line 509: Same as above.

src/igvfd/mappings/pseudobulk_set.json
```diff
@@ -423,6 +423,9 @@
                             "assay_term": {
                                 "type": "keyword"
                             },
+                            "assay_terms": {
```
zhwshen, line 426: Same as above.

src/igvfd/upgrade/file_set.py
```diff
@@ -783,3 +783,13 @@ def auxiliary_set_12_13(value, system):
     if 'barcode_map' in value:
         value['hashtag_barcode_map'] = value['barcode_map']
         del value['barcode_map']
+
+
+@upgrade_step('model_set', '6', '7')
+def model_set_6_7(value, system):
+    # https://igvf.atlassian.net/browse/IGVF-3421
+    if 'preferred_assay_titles' in value:
+        value['assay_terms'] = ['/assay-terms/OBI_0002041/']
```
zhwshen, line 792: This is somewhat shaky if there is any other preferred assay titles submitted on the prod site. Not sure what the better fix is (short of reading the whole ontology file).

zhwshen (changes requested):

src/igvfd/upgrade/file_set.py
zhwshen: Never mind. Just saw the comment in the JIRA ticket itself. It will be a 2-step upgrade with one auto-upgrade and one manual patching.

src/igvfd/mappings/analysis_set.json
shengchd: they are all mapping files - I guess it's because Mea set and Curated set have assay_term in singular, but model set has to be plural since we already have a case from SEMpl with multiple assays. But they should all have calculated property assay_titles in plural  form - which UI will use I think.

zhwshen: @khzlin How will this show up on the Facets or search config? Will people see both `assay_term` and `assay_terms`?

khzlin: this is mappings.... doesn't have any thing to do with facets or search configs.

zhwshen (dismissed):
PR looks good. Resolve conflicts before launching demo.

PR #2025, opened 2026-04-21
IGVF-3436-remove-assay-titles

shengchd:


src/igvfd/tests/fixtures/schemas/model_set.py
```diff
@@ -5,8 +5,7 @@
 def model_set_no_input(
     testapp,
     award,
-    lab,
-    software_version
```
shengchd, line 9: not relevant with the ticket, just removed it since it's no longer used

ian-whaling (approved):

PR #2026, opened 2026-04-21
IGVF-3428-zt-units

jenjou:


src/igvfd/schemas/phenotypic_feature.json
```diff
@@ -89,7 +89,8 @@
                 "milligram per deciliter",
                 "UPDRS",
                 "MMSE",
-                "SLEDAI"
+                "SLEDAI",
+                "ZT"
```
zhwshen, line 93: Do you think "ZT" also needs a description?

zhwshen (approved):

PR #2044, opened 2026-04-30
IGVF-3464-update-tab-summary

shengchd:


src/igvfd/types/file.py
```diff
@@ -132,6 +132,36 @@ def show_s3uri(externally_hosted=None):
...
 
 
+def get_transcriptome_annotation_from_reference_files(request, reference_files):
+    if not reference_files:
+        return None
+    transcriptome_annotation_set = set()
+    for ref_file in paths_filtered_by_status(request, reference_files):
+        ref_file_object = request.embed(ref_file, '@@object?skip_calculated=true')
+        if ref_file_object['content_type'] in ('transcriptome reference', 'transcriptome index'):
+            transcriptome_annotation_set.add(ref_file_object.get('transcriptome_annotation', None))
```
zhwshen, line 142: potential issue if a RefFile doesn't have assembly info. the set will be {'GRCh38', None}, making it a mixed reference. but this issue should be eventually resolved once the assembly + transcriptome ref becomes a dependency. not sure if it is worth to write a condition check of `if ref_file_object.get('transcriptome_annotation')`.

src/igvfd/types/file.py
```diff
@@ -132,6 +132,36 @@ def show_s3uri(externally_hosted=None):
...
+
+
+def get_assembly_from_reference_files(request, reference_files):
+    if not reference_files:
+        return None
+    assembly_set = set()
+    for ref_file in paths_filtered_by_status(request, reference_files):
+        ref_file_object = request.embed(ref_file, '@@object?skip_calculated=true')
+        if ref_file_object['content_type'] in ('genome reference', 'genome index'):
+            assembly_set.add(ref_file_object.get('assembly', None))
```
zhwshen, line 157: same as above.

src/igvfd/types/file.py
shengchd: we are going to leave it as is right now - since it will be corrected eventually

zhwshen (dismissed):

PR #2051, opened 2026-05-05
IGVF-1131-audit-term-name

ian-whaling:


src/igvfd/audit/ontology_term.py
```diff
@@ -33,8 +33,76 @@ def audit_ntr_term_id(value, system):
...
+            f'{detail} {audit_message.get("audit_description", "")}',
+            level=audit_message.get('audit_level', ''),
+        )
+
+
+def audit_inconsistent_ontology_term(value, system):
+    '''
+    [
+        {
+            "audit_description": "The term name is expected to match the canonical name or a synonym in the reference ontology for the given term identifier.",
```
jenjou: I think the audit description should be updated to remove "or a synonym" since the audit logic doesn't actually check the synonyms.

ian-whaling: fixed

jenjou (approved):

PR #2054, opened 2026-05-06
IGVF-3495-cls-donor-bug

zhwshen:


src/igvfd/types/file_set.py
```diff
@@ -2379,10 +2379,15 @@ def assay_slims(self, request, file_sets=None):
         }
     )
     def donors(self, request, samples=None):
+        if not samples:
```
keenangraham: I think this is equivalent to `condition='samples'` which the decorator already has? https://github.com/IGVF-DACC/igvfd/blob/54cc0e710b9adeb8252c84307e6a71eb63c702e4/src/igvfd/types/file_set.py#L2365

In other words this function will only run if there are samples.

src/igvfd/types/file_set.py
```diff
@@ -2379,10 +2379,15 @@ def assay_slims(self, request, file_sets=None):
...
+            return None
         # Filter once for non-deprecated samples
         non_deprecated_samples = paths_filtered_by_status(request, samples)
         # Filter again for non-deprecated donors
-        return paths_filtered_by_status(request, get_donors_from_samples(request, non_deprecated_samples))
+        unfiltered_donors = get_donors_from_samples(request, non_deprecated_samples)
+        # Filter donors by status (note will return [] instead of None if all deprecated donors)
+        filtered_donors = paths_filtered_by_status(request, unfiltered_donors) if unfiltered_donors else None
+        return filtered_donors or None
```
keenangraham: I think this could be replaced by:
```
if unfiltered_donors:
    return paths_filtered_by_status(request, unfiltered_donors) or None
```

src/igvfd/types/file_set.py
```diff
@@ -2379,10 +2379,15 @@ def assay_slims(self, request, file_sets=None):
         }
     )
     def donors(self, request, samples=None):
+        if not samples:
+            return None
         # Filter once for non-deprecated samples
         non_deprecated_samples = paths_filtered_by_status(request, samples)
```
keenangraham: Should this be `filtered_samples` to match `unfiltered_donors` and `filtered_donors`?

src/igvfd/types/file_set.py
```diff
@@ -2379,10 +2379,15 @@ def assay_slims(self, request, file_sets=None):
         }
     )
     def donors(self, request, samples=None):
+        if not samples:
+            return None
         # Filter once for non-deprecated samples
```
keenangraham: Kind of redundant with code.

src/igvfd/types/file_set.py
```diff
@@ -2379,10 +2379,15 @@ def assay_slims(self, request, file_sets=None):
...
         non_deprecated_samples = paths_filtered_by_status(request, samples)
         # Filter again for non-deprecated donors
```
keenangraham: Kind of redundant with code.

keenangraham (changes requested):
Maybe remove the comments, not sure they are adding much.

keenangraham (approved):

PR #2064, opened 2026-05-13
IGVF-3405-simple-samples

zhwshen:


src/igvfd/types/file_set.py
```diff
@@ -201,6 +201,95 @@ def get_preferred_assay_slims(preferred_assay_titles):
...
+    if len(term_lists) == 1:
+        return term_lists[0]
+    if len(term_lists) == 2:
+        return f'{term_lists[0]} and 1 other phenotype'
+    return f'{term_lists[0]} and {len(term_lists) - 1} more'
+
+
+def get_disease_terms_from_phenotypic_features(request, sample_object, is_pd_collection=False):
+    '''Get disease terms from phenotypic features.'''
+    disease_prefixes = ('DOID:', 'EFO:', 'HP:', 'MONDO:')
```
jenjou: I think NCIT should be valid for this as well.

src/igvfd/types/file_set.py
```diff
@@ -1126,11 +1203,21 @@ def sample_summary(self, request, samples=None):
             ):
                 # Insert the classification before the targeted_sample_term if it exists.
```
jenjou: I think it would also be helpful to add more explanation to this comment, since now it also handles putting the classification before the disease terms.

src/igvfd/types/file_set.py
```diff
@@ -1051,6 +1127,34 @@ def sample_summary(self, request, samples=None):
             if classification not in sample_classification_term_target:
                 sample_classification_term_target[classification] = set()
 
+            # Add disease terms to the sample phrase
+            disease_phrase = ''
+            disease_terms_by_sample_term = {}
+            if sample_object.get('multiplexed_samples'):
```
jenjou, line 1174: I would recommend adding a default value here `sample_object.get('multiplexed_samples', [])`, since it is possible the object doesn't have multiplexed_samples.

src/igvfd/types/file_set.py
```diff
@@ -1069,47 +1173,27 @@ def sample_summary(self, request, samples=None):
                     targeted_sample_term_object = request.embed(
                         sample_object['targeted_sample_term'], '@@object?skip_calculated=true')
                     targeted_sample_suffix = f"induced to {targeted_sample_term_object['term_name']}"
+
+                term_disease_phrase = ''
+                if sample_object.get('multiplexed_samples'):
```
jenjou, line 1227: Same comment as above.

src/igvfd/types/file_set.py
```diff
@@ -201,6 +201,96 @@ def get_preferred_assay_slims(preferred_assay_titles):
...
+    for cls_set in construct_library_sets:
+        cls_object = request.embed(cls_set, '@@object?skip_calculated=true')
+        if cls_object.get('file_set_type') == 'overexpression vector library' and cls_object.get('scope') == 'genes':
+            small_gene_list = cls_object.get('small_scale_gene_list', [])
+            if small_gene_list:
+                for gene in small_gene_list:
+                    gene_object = request.embed(gene, '@@object?skip_calculated=true')
+                    oexpr_genes.add(gene_object.get('symbol', ''))
+    oexpr_genes = sorted(list(oexpr_genes))
+    return f'overexpressing {join_multiple_terms(oexpr_genes)}' if oexpr_genes else ''
```
jenjou, line 298: Since this line is also using the helper function `join_multiple_terms()`, does that function need to be updated? It seems like it adds the "and 1 other phenotype" phrase for lists of 2 objects but in this case it is supposed to be a list of genes.

zhwshen: Removing `phenotype` and updating to join and list both elements.

jenjou (dismissed):

src/igvfd/types/file_set.py
```diff
@@ -208,6 +208,135 @@ def get_preferred_assay_slims(preferred_assay_titles):
...
+    for phrase in terms:
+        if ' induced to ' in phrase:
+            prefix, tgt = phrase.split(' induced to ', 1)
+            merged_induced.setdefault(prefix, set()).add(tgt)
+        else:
+            plain_terms.append(phrase)
+    result = plain_terms
+    for prefix, targets in merged_induced.items():
+        result.append(f'{prefix} induced to {", ".join(sorted(targets))}')
+    return result
```
jenjou, line 314: Is it possible to add more sorting here to make the returned value consistent?

I am just unsure what is better here. Do we want the entire array to be sorted as one group (thus it might be: [plain term 1], [induced term 1], [plain term 2]) or do we want all of the plain terms first followed by all of the "induced" terms?

src/igvfd/types/file_set.py
```diff
@@ -1144,49 +1296,49 @@ def sample_summary(self, request, samples=None):
...
+        taxa_rename_map = {
+            'Homo sapiens': 'human',
+            'Mus musculus': 'mouse',
+            'Saccharomyces cerevisiae': 'yeast'
+        }
+        for taxa_name, renamed_taxa in taxa_rename_map.items():
+            summary = summary.replace(taxa_name, renamed_taxa)
+
+        if summary:
+            summary = f'{summary[0].upper()}{summary[1:]}'
```
jenjou: We don't seem to have been capitalizing the summary for the regular summaries of file sets. Just wondering was this a deliberate choice that we want sample summary to always start capitalized?

zhwshen: The previous `sample_summary` starts with `taxa`, which is always capitalized. I think File set summary can be a bit different case because sometimes it is just assay names. File set description is more likely to be capitalized. So I am inclined to make it sentence case because it is somewhat of a sentence.

src/igvfd/types/file_set.py
zhwshen: I can make it sorted(result). The targets are sorted already here and each term is sorted inside the `sample_summary`, so the output will be `A induced to X, Y, Z, B induced to M`. So the sorting here would be redundant.

src/igvfd/types/file_set.py
zhwshen: Will change to lower case after discussion with Ben.

jenjou (dismissed):

PR #2068, opened 2026-05-15
IGVF-3471-matrix-file-upgrade

jenjou:


src/igvfd/schemas/changelogs/matrix_file.md
```diff
@@ -1,5 +1,43 @@
...
+* Extend `content_type` enum list to include `cell by gene matrix`.
+* Extend `content_type` enum list to include `cell by gene and guide matrix`.
+* Extend `content_type` enum list to include `cell by gene and guide and multiplexing oligo matrix`.
+* Extend `content_type` enum list to include `cell by gene and multiplexing oligo matrix`.
+* Extend `content_type` enum list to include `cell by gene and peak matrix`.
+* Extend `content_type` enum list to include `cell by guide matrix`.
+* Extend `content_type` enum list to include `cell by peak matrix`.
+* Extend `content_type` enum list to include `cell by position methylation count matrix`.
+* Extend `content_type` enum list to include `cell by UMI count matrix`.
+* Extend `content_type` enum list to include `contact matrix`.
```
shengchd: remove this line - the enum already exists

src/igvfd/upgrade/analysis_step.py
```diff
@@ -215,3 +215,45 @@ def analysis_step_10_11(value, system):
...
+def analysis_step_11_12(value, system):
+    # https://igvf.atlassian.net/browse/IGVF-3471
+    notes = value.get('notes', '')
+    old_input_content_types = value.get('input_content_types', [])
+    new_input_content_types = []
+    old_output_content_types = value.get('output_content_types', [])
+    new_output_content_types = []
+    upgrade_map = {
+        'allele specific sparse gene count matrix': 'allele specific cell by gene matrix',
+        'annotated multimodal CRISPR matrix': 'cell by gene matrix',
```
shengchd, line 230: is it intentional this map from 'annotated multimodal CRISPR matrix' is not included in file upgrade code?

src/igvfd/tests/test_upgrade_analysis_step.py
```diff
@@ -52,3 +52,13 @@ def test_analysis_step_upgrade_10_11(upgrader, analysis_step_v10):
...
     assert set(value['output_content_types']) == {'global differential expression', 'local differential expression'}
     assert value['schema_version'] == '11'
+
+
+def test_analysis_step_upgrade_11_12(upgrader, analysis_step_v11):
+    value = upgrader.upgrade('analysis_step', analysis_step_v11, current_version='11', target_version='12')
+    assert len(value['input_content_types']) == 2
+    assert set(value['input_content_types']) == {'cell by gene matrix', 'kallisto cell by gene matrix'}
+    assert len(value['output_content_types']) == 3
+    assert set(value['output_content_types']) == {'annotated cell by peak matrix',
```
shengchd, line 62: can you add assert for notes?

src/igvfd/upgrade/file.py
```diff
@@ -779,3 +779,62 @@ def reference_file_23_24(value, system):
...
+def matrix_file_9_10(value, system):
+    # https://igvf.atlassian.net/browse/IGVF-3471
+    notes = value.get('notes', '')
+
+    principal_dimension = value.get('principal_dimension', '')
+    secondary_dimensions = ', '.join(sorted(value.get('secondary_dimensions', [])))
+    content_type = value.get('content_type', '')
+    filtered = value.get('filtered', None)
+
+    map = {
```
shengchd: minor point: might better rename map to upgrade_map as in AS, since map is a built-in function in python

shengchd (changes requested):

src/igvfd/upgrade/analysis_step.py
jenjou: Yes. I could not find any examples of `annotated multimodal CRISPR matrix`  files so I decided to leave it out for files. If the file upgrade code is working correctly, then it should get replaced with the placeholder value `spot by gene matrix` (which is the default for anything that doesn't have a specific mapping).

Analysis Step is simpler and I don't need to consider the dimensions, so it seemed faster to just list all the content types, even if it is going to map to a default placeholder.

shengchd: that makes sense, was just curious about the intention.

shengchd (dismissed):

PR #2073, opened 2026-05-20
IGVF-3466-reintroduce-readout

ian-whaling:


src/igvfd/upgrade/ontology_term.py
```diff
@@ -330,3 +330,29 @@ def assay_term_21_22(value, system):
...
+    preferred_assay_titles = value.get('preferred_assay_titles', [])
+    if not preferred_assay_titles:
+        return
+    if any(title in REMOVED_CRISPR_TILING_PREFERRED_ASSAY_TITLES for title in preferred_assay_titles):
+        notes = value.get('notes', '')
+        removed_list = ', '.join(
+            title for title in preferred_assay_titles
+            if title in REMOVED_CRISPR_TILING_PREFERRED_ASSAY_TITLES
+        )
+        value['preferred_assay_titles'] = ['CRISPR FACS screen']
```
jenjou: This is an edge case, but shouldn't we retain the existing preferred assay titles that are not one of the "tiling screen" ones? This seems to overwrite everything to be CRISPR FACS screen even if only 1 of the existing titles needs to be removed.

jenjou (dismissed):

PR #2074, opened 2026-05-20
IGVF-3502-tar-image-file

jenjou:


src/igvfd/schemas/tabular_file.json
```diff
@@ -395,6 +399,7 @@
                 "per cell type enrichment scores",
                 "pipeline parameters",
                 "plus strand modification state",
+                "post-segmentation transcript and cell locations",
```
zhwshen, line 402: Do you need to add any of these terms to the TabularFile audit for missing genome or transcriptome reference files?

src/igvfd/schemas/tabular_file.json
```diff
@@ -417,6 +422,9 @@
                 "reporter genomic variant effects",
                 "sample sort parameters",
                 "selected normalized variants profile",
+                "spatial transcriptomics cell metadata",
```
zhwshen, line 425: Same as above.

src/igvfd/schemas/tabular_file.json
jenjou: Yes, I notice there are other terms missing from the audit, so I will just do a general update and add terms we have missed previously

zhwshen (dismissed):

zhwshen (approved):

PR #2076, opened 2026-05-22
IGVF-3510-add-donor-gene-matrix

khzlin:


src/igvfd/schemas/changelogs/matrix_file.md
```diff
@@ -2,6 +2,7 @@
 
 ### Schema version 10
 
+* Extend `content_type` enum list to include `donor by gene matrix`.
```
shengchd, line 5: need to be in a separate section

shengchd (changes requested):

zhwshen (dismissed):

PR #2085, opened 2026-06-01
IGVF-3489-fix-pseudobulk-dependency

khzlin:


src/igvfd/types/file_set.py
```diff
@@ -2814,3 +2789,23 @@ def assay_slims(self, request, input_file_sets=None):
...
+                'title': 'Donor',
+                'description': 'Donor of a sample associated with this analysis set.',
+                'type': 'string',
+                'linkTo': 'Donor'
+            },
+            'notSubmittable': True,
+        }
+    )
+    def donors(self, request, samples=None):
+        return get_donors_from_samples(request, samples)
```
zhwshen: Will need additional codes:
1) `get_donors_from_samples` does not filter sample or donor status. Could have deleted samples or samples' deleted donors in the calculated list. Consider doing this https://github.com/IGVF-DACC/igvfd/blob/0689d8333802f43fa6011e77bd694f358c9cd51b/src/igvfd/types/file_set.py#L2430 to filter for only active donors.

2) Need a test to check this calculation.

zhwshen (changes requested):

zhwshen (dismissed):

zhwshen (dismissed):

PR #2086, opened 2026-06-01
WRN-1026-stuff to make docs

hitz:
No need to actually merge this it's just for review


igvf_field_glossary.json
```diff
@@ -0,0 +1,917 @@
...
+        "VAMP-seq (MultiSTEP)": "Variant of VAMP-seq that measures the effects of missense variants on secretion and post-translational modification of secreted proteins using a cell surface display method.",
+        "varACCESS": "A high-throughput assay that integrates genome variants into chromatin accessibility analysis.",
+        "Variant painting via fluorescence": "A modification of Cell Painting that employs fluorescent protein fusions to protein coding variants to characterize localization alongside cellular morphology.",
+        "Variant painting via immunostaining": "A modification of Cell Painting that employs epitope tagging and immunostaining of protein coding variants to characterize localization alongside cellular morphology.",
+        "Variant-EFFECTS": "A CRISPR-based screen that combines CRISPR prime editing, RNA fluorescence in situ hybridization (FlowFISH), and cell sorting to quantitatively measure the effects of designed DNA sequence variants on gene expression.",
+        "WGS": "Whole Genome Sequencing (WGS) is an assay to determine the entire sequence of an organism's genome."
+    },
+    "software": {
```
ian-whaling, line 83: You will need to retrieve `title` instead of `name` since that is the field that is used to write to softwares on files_filesets.
https://github.com/IGVF-DACC/igvf-catalog/blob/dev/data/adapters/file_fileset_adapter.py#L246

You might want to give the agent writing these scripts the context of the file_fileset_adapter.py and files_filesets schema because that is the collection that you will need to map from for definitions.

igvf_field_glossary.json
```diff
@@ -0,0 +1,917 @@
...
+            "variant drug annotations": "Annotations describing relationships between variants and drug response.",
+            "variant effects": "The genetic variant effects on gene expression.",
+            "variant functions": "Predictions of the functional effects of genetic variants.",
+            "variant localization impacts": "Output of a variant painting experiement describing the measured impact of variants.",
+            "variant pathogenicity": "A file containing variant pathogenicity scores.",
+            "variant to cluster mapping": "A file that maps variants to specific morphological \"clusters\" or groups of similar phenotypes.",
+            "variant to element mapping": "A file describing the mapping of genetic variants to genomic elements.",
+            "variants": "A table listing genetic variants, which may be called from IGVF data or collected from external resources."
+        },
+        "content_type_fields": {
```
ian-whaling, line 617: `content_type` is not percolated anywhere to the catalog DB, so I would not rely on it.
Also `file_format_specifications` are not very useful as those fields are mutated in the downstream DB. They may be useful for writing definitions for the schema fields. The schema fields are actually representative of the record fields in the DB.

Then we will need to tie in any mutation of field name between DB and API and harmonization of the same field used between two methods of the same collection that have distinct definitions for that field.
In short we need to write definitions for DB and (potentially) API.

igvf_field_glossary.json
```diff
@@ -0,0 +1,917 @@
...
+                "spdi": "Variant identifier in SPDI format",
+                "SPDI_ID": "Variant identifier in SPDI notation (Sequence:Position:Deleted:Inserted)",
+                "VariantChr": "Chromosome where the variant is located (e.g., chr1)",
+                "VariantEnd": "Genomic end position of the variant (same as start for SNVs)",
+                "VariantStart": "Genomic start position of the variant (1-based)",
+                "{organ}.TLand": "51 columns each representing a predicted functional variant probability score for {organ}"
+            }
+        }
+    },
+    "biogrid_experimental_systems": {
```
ian-whaling, line 838: I am not sure what these are

igvf_field_glossary_curated.json
```diff
@@ -0,0 +1,789 @@
...
+            "Thermal Shift Assay": "An interaction is demonstrated between a protein and a ligand by measuring the change in a protein's thermal stability upon ligand binding. This shift can be measured by nano-differential scanning fluorimetry (Nano-DSF) (e.g.PMID: 36650907, Fig. 2 and 3), differential scanning calorimetry (DSC) (e.g. PMID: 32828280, Fig. 1D) or other methods.",
+            "Two-hybrid": "An interaction is inferred when a bait protein is expressed as a DNA binding domain (DBD) fusion, a prey protein is expressed as a transcriptional activation domain (TAD) fusion and the interaction is measured by reporter gene activation (e.g. PMID: 9082982, Table 1)."
+        },
+        "genetic_interactions": {
```
ian-whaling, line 733: or these

ian-whaling (changes requested):
The same comments I left on igvf_field_glossary.json should be applied to igvf_filed_glossary_curated.json.

- `content_type` enum_descriptions are not useful because the `content_type` do not actually appear anywhere in catalog DB.
- file format specification definitions are not useful because the field names are mutated at both DB load and API call. 
- software should use `title` not `name` for mapping.

encode_set.json
```diff
@@ -0,0 +1,183 @@
+{
```
ian-whaling, line 1: what is this for

hitz: nothing a temporary file, could remove it

igvf_field_glossary.json
hitz: they are the sub-methods from biogrid for protein-protein interactions (and genetic interactions)

igvf_field_glossary.json
hitz: can you get met a list of fields?

igvf_field_glossary_curated.json
hitz: yeah from BioGrid.  Harmless at worse.

PR #2089, opened 2026-06-03
IGVF-3488-restrict-inclusion-list

shengchd:


src/igvfd/schemas/tabular_file.json
```diff
@@ -175,7 +175,6 @@
                                 "element to gene interactions",
                                 "elements reference",
                                 "eQTL",
-                                "exclusion list regions",
```
zhwshen, line 178: Should also check the long exclusion/inclusion list for requiring reference files under TabFile audit.

zhwshen (changes requested):

src/igvfd/schemas/tabular_file.json
shengchd: checked, I don't see this enum listed under TabFile audit: https://github.com/IGVF-DACC/igvfd/blob/0c8d444342c51f22316c5048cb7d994462477822/src/igvfd/audit/file_set.py#L1196

zhwshen (approved):

PR #2093, opened 2026-06-05
IGVF-3525-transposase

ian-whaling:


src/igvfd/searches/configs/analysis_set.py
```diff
@@ -108,6 +108,12 @@ def analysis_set():
                 'description': 'Treatments applied to the sample with the purpose of perturbation.',
                 'optional': True
             },
+            'samples.construct_delivery_methods': {
+                'title': 'Construct Delivery Method',
```
zhwshen: Method or Method**s**?

src/igvfd/searches/configs/measurement_set.py
```diff
@@ -101,6 +101,12 @@ def measurement_set():
                 'description': 'Treatments applied to the sample with the purpose of perturbation.',
                 'optional': True
             },
+            'samples.construct_delivery_methods': {
+                'title': 'Construct Delivery Method',
```
zhwshen: Same as above.

zhwshen (changes requested):

src/igvfd/searches/configs/measurement_set.py
ian-whaling: fixed

src/igvfd/searches/configs/analysis_set.py
ian-whaling: fixed

zhwshen (approved):

PR #2101, opened 2026-06-10
IGVF-3544-dedup-samp-sum

zhwshen:


src/igvfd/types/file_set.py
```diff
@@ -1290,7 +1295,7 @@ def sample_summary(self, request, samples=None, donors=None, construct_library_s
                         )
                     else:
                         terms_by_classification = f'{terms_by_classification} {classification}'
-            elif any(x in terms_by_classification for x in [
+            elif classification in terms_by_classification or any(x in terms_by_classification for x in [
```
jenjou: I think this logic could be slightly simplified as it seems like any thing that is true for the 2nd condition `any(x in terms_by_classification for x in ['differentiated cell specimen', 'reprogrammed cell specimen', 'pooled cell specimen', 'primary cell']` must also be true for the first condition `classification in terms_by_classification`.

src/igvfd/types/file_set.py
```diff
@@ -1269,7 +1274,7 @@ def sample_summary(self, request, samples=None, donors=None, construct_library_s
             # Differentiated, reprogrammed, pooled cell specimen can be merged
             # into the terms_by_classification before this. Therefore we don't
             # want to append it to the terms_by_classification a second time.
-            elif not any(x in terms_by_classification for x in [
+            elif classification not in terms_by_classification and not any(x in terms_by_classification for x in [
```
jenjou: Same comment as below. I think if condition 2 is true then condition 1 is also true here.

src/igvfd/types/file_set.py
zhwshen: I can simplify the expression, but I’d keep both parts because the second check covers normalized specimen phrases that are not always exact matches of classification. Might be a better approach to explicitly spell out the conditions.

src/igvfd/types/file_set.py
zhwshen: Same as above.

jenjou (approved):

PR #2104, opened 2026-06-11
IGVF-3533-pseu-set-summary-update

jenjou:


src/igvfd/types/file_set.py
```diff
@@ -2806,27 +2806,34 @@ class PseudobulkSet(FileSet):
...
+        if merged:
+            merged_phrase = 'merged'
+        assay_phrase = ''
+        if preferred_assay_titles:
+            assay_phrase = ', '.join(sorted(preferred_assay_titles))
+        taxa_phrase = ''
+        if donors:
+            taxa = set()
+            for donor in donors:
+                donor_object = request.embed(donor, '@@object')
```
zhwshen: Is it less calculation if it is `@@object?skip_calculated=true`?

zhwshen (dismissed):

PR #2108, opened 2026-06-12
IGVF-3519-crispr-screen-biometric

ian-whaling:


src/igvfd/upgrade/file_set.py
```diff
@@ -914,3 +914,11 @@ def file_set_33_34(value, system):
...
         notes += f'This file set previously used perturb-SHARE-seq as a preferred_assay_titles, but it has been updated to MORF-SHARE-seq via an upgrade.'
         value['notes'] = notes.strip()
+
+
+@upgrade_step('measurement_set', '46', '47')
+def measurement_set_46_47(value, system):
+    # https://igvf.atlassian.net/browse/IGVF-3519
+    if 'crispr_readout' in value:
+        value['crispr_screen_readout'] = value['crispr_readout']
+        del value['crispr_readout']
```
zhwshen, line 924: Do we want an upgrade note here?

src/igvfd/tests/test_upgrade_measurement_set.py
```diff
@@ -385,3 +385,10 @@ def test_measurement_set_upgrade_45_46(upgrader, measurement_set_v45):
     value = upgrader.upgrade('measurement_set', measurement_set_v45, current_version='45', target_version='46')
     assert value['schema_version'] == '46'
     assert value['preferred_assay_titles'] == ['MORF-SHARE-seq']
+
+
+def test_measurement_set_upgrade_46_47(upgrader, measurement_set_v46):
+    value = upgrader.upgrade('measurement_set', measurement_set_v46, current_version='46', target_version='47')
+    assert value['schema_version'] == '47'
+    assert value['crispr_screen_readout'] == 'gRNA sequencing'
+    assert 'crispr_readout' not in value
```
zhwshen, line 394: If decided to add `note`, would be nice to check that too.

src/igvfd/types/file_set.py
```diff
@@ -22,6 +22,8 @@
     '/assay-terms/NTR_0001101/',  # in vivo CRISPR screen using single cell RNA-seq
 }
 
+PROTEIN_ABUNDANCE_TERM_NAME = 'protein abundance'
```
zhwshen, line 25: Is this better as either a string or an array of the protein abundance term IDs?

src/igvfd/tests/test_audit_measurement_set.py
```diff
@@ -267,7 +267,135 @@ def test_audit_missing_crispr_readout(
...
+            'preferred_assay_titles': ['CRISPR FACS screen'],
+            'control_types': ['wildtype'],
+        }
+    )
+    res = testapp.get(measurement_set['@id'] + '@@audit')
+    assert all(
+        error['category'] != 'missing targeted genes'
+        for error in res.json['audit'].get('NOT_COMPLIANT', [])
+    )
+    assert all(
```
zhwshen: Is this a left over code from before as this audit was upgraded from WARNING to NON_COMPLIANT?

zhwshen (changes requested):

src/igvfd/types/file_set.py
ian-whaling: the property is an enum of strings

src/igvfd/tests/test_audit_measurement_set.py
ian-whaling: fixed

src/igvfd/upgrade/file_set.py
ian-whaling: no, doesn't matter the field names are synonymous

src/igvfd/types/file_set.py
zhwshen: My typo. I meant is it more accurate to use the term ID as the string value instead of term name?

zhwshen: It will be an NTR, so yes, term name is better.

zhwshen (dismissed):

zhwshen (dismissed):

PR #2112, opened 2026-06-17
IGVF-3561-sce2g-cell-annot

zhwshen:


src/igvfd/schemas/mixins.json
```diff
@@ -1110,5 +1110,26 @@
                 "igvf_utils": true
             }
         }
+    },
+    "cell_annotation_properties": {
+        "cell_type": {
+            "title": "Cell Type",
+            "description": "The ontology term that describes the cell type of the cells in this pseudobulk.",
```
jenjou: Should this description be updated? In the case of the scE2G predictions, they do use pseudobulks as input, so perhaps it is OK as is.

zhwshen: @jenjou @shengchd I can update this to be "of this pseudobulk and/or prediction". How does this sound?

jenjou: Ok with me.

jenjou (dismissed):

src/igvfd/schemas/mixins.json
```diff
@@ -1110,5 +1110,26 @@
...
+            "type": "string",
+            "linkTo": "SampleTerm",
+            "submissionExample": {
+                "appscript": "[\"/sample-terms/CL_1230000/\"]",
+                "igvf_utils": "/sample-terms/CL_1230000/"
+            }
+        },
+        "cell_qualifier": {
+            "title": "Cell Qualifier",
+            "description": "A qualifier that provides additional detail about the cell type annotation or the source biosample.",
```
jenjou: In the description, can you change "source biosample" to "parent sample"? This will keep it more consistent with other properties that refer to parent samples.

jenjou (approved):

PR #2113, opened 2026-06-18
IGVF-2994-add-batch-download-filters

keenangraham:


src/igvfd/metadata/metadata.py
```diff
@@ -79,6 +79,10 @@ def file_satisfies_inequality_constraints(file_, positive_file_inequalities):
     return True
 
 
+def file_not_uploaded(file_):
+    return file_['upload_status'] in ('pending', 'file not found')
```
ottojolanki, line 83: Do we want to include `'invalidated'` as well?

ottojolanki (dismissed):

ottojolanki (approved):

PR #2118, opened 2026-06-19
IGVF-3538-assays-from-CLS

ian-whaling:


src/igvfd/types/file_set.py
```diff
@@ -121,6 +121,45 @@ def get_file_set_props_for_summary_and_samples(request, file_set):
...
 
 
+def get_assay_contributing_input_file_sets(request, input_file_sets):
+    '''Return input file sets that contribute assay metadata to analysis-derived file sets.
+
+    Construct library sets are assay-agnostic and should not contribute assay information
+    when other input file sets are present. Analysis sets comprised solely of construct
+    library sets are an exception and inherit assays from those construct library sets.
+    When other input file sets are present, upstream analysis sets that are comprised
+    solely of construct library sets are also excluded.
```
shengchd: so if the input file sets only contain analysis set, that comprises solely CLS, should this function return that input analysis set? (think right now it's skipped)

shengchd (commented):
code is good, just have a question on calculation logic

src/igvfd/types/file_set.py
ian-whaling: Don't think this is a real case because in that sense one should just submit the data to the same analysis set, but I added a case for it. That analysis set should inherit assay metadata.

shengchd (dismissed):

src/igvfd/types/file_set.py
```diff
@@ -127,6 +127,98 @@ def get_file_set_props_for_summary_and_samples(request, file_set):
...
+        cls_inputs = [
+            fileset for fileset in input_file_sets
+            if fileset.startswith('/construct-library-sets/')
+        ]
+        if cls_inputs:
+            return cls_inputs
+        return [
+            fileset for fileset in input_file_sets
+            if curated_set_has_assay_metadata(fileset)
+        ]
```
shengchd, line 211: try to understand what supposed to be returned here, this will always return [], since inputs_are_cls_only_equivalent can only be true if no input curated set has assay metadata

src/igvfd/types/file_set.py
```diff
@@ -127,6 +127,98 @@ def get_file_set_props_for_summary_and_samples(request, file_set):
...
+                if curated_set_has_assay_metadata(fileset):
+                    return False
+                continue
+            if fileset.startswith('/analysis-sets/'):
+                analysis_object = request.embed(
+                    fileset,
+                    '@@object_with_select_calculated_properties?field=input_file_sets'
+                )
+                upstream_inputs = analysis_object.get('input_file_sets', [])
+                if not upstream_inputs or not inputs_are_cls_only_equivalent(upstream_inputs):
```
shengchd: risk for recursion here since the function calls itself (input of A is B, input of B is A, though technically not valid case), can consider add inspected_filesets like analysis set summary part to prevent that recursion error happening

src/igvfd/types/file_set.py
ian-whaling: good catch you're right this would return False/[] as soon as any curated set has assay metadata.

src/igvfd/types/file_set.py
ian-whaling: fixed

src/igvfd/types/file_set.py
ian-whaling: fixed + Wrote a test

src/igvfd/types/file_set.py
ian-whaling: can't actually test recursions in test environment so didn't create a test

shengchd (dismissed):

PR #2119, opened 2026-06-22
IGVF-3545-depmap-regex

khzlin:


src/igvfd/upgrade/donor.py
```diff
@@ -202,3 +202,13 @@ def donor_14_15(value, system):
     # https://igvf.atlassian.net/browse/IGVF-1804
     if 'phenotypic_features' in value:
         del value['phenotypic_features']
+
+
+@upgrade_step('human_donor', '15', '16')
+def human_donor_14_15(value, system):
```
shengchd: typo in version

src/igvfd/upgrade/donor.py
```diff
@@ -202,3 +202,13 @@ def donor_14_15(value, system):
...
+    # https://igvf.atlassian.net/browse/IGVF-3545
+    if 'dbxrefs' in value:
+        value['dbxrefs'] = [
+            dbxref for dbxref in value['dbxrefs']
+            if not dbxref.startswith('DepMap:')
```
shengchd, line 213: do you want to remove all DepMap cases?

shengchd (changes requested):

src/igvfd/upgrade/donor.py
shengchd: Also need to consider if the list is empty after removing DepMap cases - delete it then

src/igvfd/upgrade/donor.py
khzlin: thanks.

src/igvfd/upgrade/donor.py
khzlin: yeah, because DepMap:ACH- is samples, but DepMap:PT- is for donor. We haven't patched anything anywhere so I think it is safe to just remove

shengchd (approved):

PR #2123, opened 2026-06-23
IGVF-3562-version-ref-file

shengchd:


src/igvfd/types/file_set.py
```diff
@@ -1662,6 +1662,31 @@ def transcriptome_annotations(self, request, files=None):
...
+        }
+    )
+    def versions(self, request, files=None):
+        version_values = set()
+        files = paths_filtered_by_status(request, files)
+        if files:
+            for current_file_path in files:
+                file_object = request.embed(
+                    current_file_path, '@@object_with_select_calculated_properties?field=version&field=@type&field=status')
+                if 'ReferenceFile' in file_object.get('@type', []) and file_object.get('version') and file_object.get('status') != 'archived':
```
khzlin, line 1686: why !='archived' ? do you mean not deleted?

shengchd: I want to exclude archived files for cases e.g. https://data.igvf.org/curated-sets/IGVFDS7295JLJQ/, we archived the older version in the same curated set, so versions calculation only get the most recent (released) one.

khzlin: I see.

khzlin (dismissed):

PR #2145, opened 2026-07-09
IGVF-3461-drop-down-facet

khzlin:


forresttanaka (dismissed):
With the note that neither of us know what the facets.py file is for. I doubt this causes a problem, as it doesn't cause a problem today.

forresttanaka (approved):

PR #2147, opened 2026-07-13
IGVF-3329-cell-line-donor-audit

jenjou:


src/igvfd/audit/in_vitro_system.py
```diff
@@ -58,8 +58,40 @@ def audit_cell_fate_change_protocol_document_type(value, system):
             yield AuditFailure(audit_message.get('audit_category', ''), f'{detail} {audit_message.get("audit_description", "")}', level=audit_message.get('audit_level', ''))
 
 
+def audit_inconsistent_cell_line_donor(value, system):
+    '''
+    [
+        {
+            "audit_description": "In vitro systems of established cell lines are expected to link to the official IGVF donor for that cell line.",
+            "audit_category": "inconsistent cell line donor",
+            "audit_level": "WARNING"
```
ian-whaling: why not an error

src/igvfd/audit/in_vitro_system.py
```diff
@@ -58,8 +58,40 @@ def audit_cell_fate_change_protocol_document_type(value, system):
...
+        }
+    ]
+    '''
+    audit_message = get_audit_message(audit_inconsistent_cell_line_donor)
+    sample_term_object = system.get('request').embed(value['sample_terms'][0] + '@@object?skip_calculated=true')
+    sample_term_name = sample_term_object.get('term_name', '')
+    sample_term_dbxrefs = sample_term_object.get('dbxrefs', [])
+    if any([x.startswith('Cellosaurus') for x in sample_term_dbxrefs]):
```
ian-whaling: This necessitates that we patch `dbxrefs` onto cell line ontologies, do you have concern with that

ian-whaling: could it be CLO: or x.startswith('Cellosaurus')?

jenjou: I considered using the sample term prefix but we do have both CLO and EFO for cell lines. While CLO is pretty much cell line specific I think EFO is more broad and could include some things we don't want to audit. (Example, `/sample-terms/EFO_0004905/` is for induced pluripotent stem cell which shouldn't be checked.)

src/igvfd/audit/in_vitro_system.py
jenjou: I could increase the severity, wasn't sure what we wanted it at.

ian-whaling (approved):

PR #2148, opened 2026-07-13
Wrn 1026 json field docs

hitz:
ok took out the content types and field definitions for now


ian-whaling (changes requested):
The software needs to be the software `title` not the software `name`.
That is the field that is being written to files_filesets.
https://github.com/IGVF-DACC/igvf-catalog/blob/13aa444f6b988ae41c9e017b7e40b54ed6f3ad6d/data/adapters/file_fileset_adapter.py#L246

PR #2152, opened 2026-07-16
IGVF-3603-parse-tap-seq

ian-whaling:


src/igvfd/schemas/changelogs/curated_set.md
```diff
@@ -2,6 +2,7 @@
 
 ### Minor changes since schema version 11
 
+* Extend `preferred_assay_titles` enum list to include `Parse TAP-seq`.
```
zhwshen, line 5: Curated Set also has preferred assay titles.

zhwshen (changes requested):
Preferred assay titles are also on ModelSet.

zhwshen (approved):

PR #2159, opened 2026-07-21
IGVF-3615-scntseq-content-type

zhwshen:


shengchd (changes requested):
Also need to update AS changelog

shengchd (approved):

PR #2164, opened 2026-07-25
IGVF-3614-encode-measurement

ian-whaling:


src/igvfd/schemas/measurement_set.json
```diff
@@ -432,6 +432,19 @@
...
                 "igvf_utils": "10X Chromium Single Cell 5 prime v3"
             }
+        },
+        "dbxrefs": {
+            "comment": "Supported external resources include: GEO, MaveDB, ENCODE.",
+            "items": {
+                "title": "External identifier",
+                "description": "Identifier from an external resource that may have 1-to-1 or 1-to-many relationships with IGVF file sets.",
+                "type": "string",
+                "pattern": "^(GEO:GSE\\d+|urn:mavedb:\\d{8}-[0a-z]+-\\d+|ENCODE:.+)$"
```
jenjou: Should the regex for ENCODE be more strict to exactly match an ENCODE accession, or do we not require that?

ian-whaling: I can make it specific to experiments, i.e. ENCSR* if that is what you are proposing

ian-whaling: But I did just copy this from the existing regex in reference files

jenjou (approved):

PR #2165, opened 2026-07-25
IGVF-3530-differential-expression

ian-whaling:


src/igvfd/upgrade/file.py
```diff
@@ -941,3 +941,26 @@ def signal_file_14_15(value, system):
...
+@upgrade_step('tabular_file', '25', '26')
+def tabular_file_25_26(value, system):
+    # https://igvf.atlassian.net/browse/IGVF-3530
+    notes = value.get('notes', '')
+    upgrade_map = {
+        'global differential expression': 'global differential expression per element',
+        'local differential expression': 'global differential expression per element',
+        'cis differential expression quantifications per guide': 'local differential expression per guide',
+        'trans differential expression quantifications per guide': 'global differential expression per guide',
+        'cis differential expression quantifications per element': 'local differential expression per element',
```
shengchd, line 955: just to confirm if this mapping is correct, ticket description has (might be a typo):
cis differential expression quantifications per element
Rename “cis” to “global” and drop “quantifications”

ian-whaling: Yeah description is wrong, should be cis --> local, trans --> global as is done in the upgrade

ian-whaling: fixed the ticket description

src/igvfd/upgrade/file.py
```diff
@@ -941,3 +941,26 @@ def signal_file_14_15(value, system):
...
         value['notes'] = notes.strip()
+
+
+@upgrade_step('tabular_file', '25', '26')
+def tabular_file_25_26(value, system):
+    # https://igvf.atlassian.net/browse/IGVF-3530
+    notes = value.get('notes', '')
+    upgrade_map = {
+        'global differential expression': 'global differential expression per element',
+        'local differential expression': 'global differential expression per element',
```
ian-whaling: Also just realized this should be local differential expression per element

shengchd (dismissed):

PR #2177, opened 2026-07-31
IGVF-3624-spatial-gene-pred

shengchd:


zhwshen (changes requested):
This content type also needs to be added to the exclusion list for genome reference. https://github.com/IGVF-DACC/igvfd/blob/b3a40725fea358684c4630685eaf921efa99c936/src/igvfd/audit/file_set.py#L1266

zhwshen (approved):

PR #2184, opened 2026-08-06
IGVF-3632-inconsistent-sample-catlas

shengchd:


src/igvfd/tests/test_audit_pseudobulk_set.py
```diff
@@ -88,6 +88,25 @@ def test_audit_pseudobulk_set_sample_matches_input(
...
+    # own and so should be exempt from this audit.
+    testapp.patch_json(
+        pseudobulk_set_base['@id'],
+        {'input_file_sets': [curated_set_external_sequencing['@id']]}
+    )
+    res = testapp.get(pseudobulk_set_base['@id'] + '@@audit')
+    assert all(
+        error['category'] != 'inconsistent samples'
+        for error in res.json['audit'].get('ERROR', [])
+    )
```
khzlin, line 102: can you add test for when 'inconsistent samples' is triggered?

khzlin (changes requested):

src/igvfd/tests/test_audit_pseudobulk_set.py
shengchd: I realized I should put test inside an existing function which has test case when it's triggered.

khzlin (dismissed):

PR #2192, opened 2026-08-12
IGVF-3647-generalize-demuxd

ian-whaling:


src/igvfd/audit/analysis_set.py
```diff
@@ -29,77 +29,60 @@ def get_assay_terms(value, system):
 def audit_analysis_set_multiplexed_samples(value, system):
     '''
     [
-        {
-            "audit_description": "Analysis sets are only expected to specify a demultiplexed sample if it has samples and they are all multiplexed.",
-            "audit_category": "unexpected demultiplexed sample",
-            "audit_level": "WARNING"
-        },
```
ian-whaling, line 36: I just removed this audit coz I do not think it can be replaced with something else

zhwshen (approved):

PR #2200, opened 2026-08-17
IGVF-3659-fix-audit-onlist

ian-whaling:


src/igvfd/audit/measurement_set.py
```diff
@@ -1095,7 +1095,8 @@ def audit_inconsistent_onlist_files(value, system):
     '''
     KIT_TO_ONLIST = {
         '10X Chromium Single Cell Multiome ATAC + Gene Expression': {
-            'e3d4c9b177b3ef177c90363bca8efd61': '3M-february-2018.txt.gz'
+            '91f5bd173373fa1815830444480236fb': 'atac_737K-arc-v1.txt.gz',
```
zhwshen, line 1098: I think that it is better to do an `assay: onlist` check in this case, so it can check cases where people accidentally put down RNA onlist for ATAC. All other kits have a single onlist, except this one that is by assay type. This has happened before for single cell.

zhwshen (changes requested):

src/igvfd/audit/measurement_set.py
ian-whaling: I did not want to make the audit any more complicated by introducing an assay check. There is only one Multiome Perturb-seq dataset from Norman lab and I do not expect there to be many more

Other kits do allow multiple onlists btw

zhwshen: It is for a small submission, OK for pass.

zhwshen (dismissed):

PR #2202, opened 2026-08-18
IGVF-3663-transcriptome-34

ian-whaling:


src/igvfd/audit/file.py
```diff
@@ -168,6 +168,7 @@ def audit_file_mixed_assembly_transcriptome_annotation(value, system):
             'GENCODE 24',
             'GENCODE 28',
             'GENCODE 32',
+            'GENCODE 34',
```
shengchd, line 171: can you add GENCODE 38 as well? somehow missed before

shengchd (commented):
code is good, a request to fix the previous bug

src/igvfd/audit/file.py
ian-whaling: fixed

shengchd (dismissed):

PR #2205, opened 2026-08-20
IGVF-3522-varaccess-enums

jenjou:


ian-whaling (changes requested):
Can you also add VarACCESS to reporter assay preferred_assay_slims?

ian-whaling (dismissed):

